PR #14, opened 2025-04-15
Moved Simulation code into its own separate module

ColeFMeldorf:


ASP_tests.py
```diff
@@ -31,22 +38,89 @@ def test_SNID_to_loc():
...
+    band = 'F184'
+    roman_bandpasses = galsim.roman.getBandpasses()
+    lam = 1293  # nm
+    sed = galsim.SED(galsim.LookupTable([100, 2600], [1, 1],
+                     interpolant='linear'), wave_type='nm',
+                     flux_type='fphotons')
+    sim_psf = \
+        galsim.ChromaticOpticalPSF(lam, diam=2.36, aberrations=galsim.roman.
+                                   getPSF(1, band, pupil_bin=1).aberrations)
+    convolved = simulate_galaxy(bg_gal_flux=9e5, deltafcn_profile=False,
```
wmwv, line 88: Is there a random seed you need to control here to make sure the noise realization is the same for `simulate_galaxy` and `simulate_supernovae`?

wmwv (changes requested):
Thanks for adding the tests.  One quick question about random seeds to ensure matching image arrays.

PR #17, opened 2025-04-21
Reintroduced Error Calculation

ColeFMeldorf:


ColeFMeldorf:
Apologies, I setup my codespace to automatically delete trailing whitespace, so there are a ton of unimportant changes in the git diff. The only significant change is reintroducing the code that calculates error at line 375 of RomanASP.py, and that code is now saved as a part of the lightcurve file in AllASPFuncs.

RomanASP.py
```diff
@@ -396,15 +408,17 @@ def main():
...
+        # Saving the output. The output needs two sections, one where we
+        # create a lightcurve compared to true values, and one where we save
+        # the images.
 
-        #Saving the output. The output needs two sections, one where we create a lightcurve compared to true values, and one where we save the images.
-        
         if use_real_images:
             identifier = str(ID)
-            lc = build_lightcurve(ID, exposures, sn_path, confusion_metric, detim, X, use_roman, band, object_type)
+            lc = build_lightcurve(ID, exposures, sn_path, confusion_metric, \
```
wmwv, line 417: I would suggest taking the opportunity here to calculate flux here, just like you do sigma_flux and then pass each to `build_lightcurve`.  It's a little non-parallel to have `X`, which `build_lightcurve` processes, and then `sigma_flux`, which `build_lightcurve` just accepts.

This will also simplify the calling sequence as you won't need to path `detim`, which I think is only used in `build_lightcurve` to specify the number of images to index lightcurve flux points from `X`.

wmwv (changes requested):
Consider re-structuring `build_lightcurve` and `build_lightcurve_sim` slightly to just accept `flux`, in parallel with `sigma_flux`.

RomanASP.py
```diff
@@ -371,6 +371,7 @@ def main():
             X, istop, itn, r1norm = lsqr[:4]
             print(istop, itn, r1norm)
 
+        flux = X[-detim:]
```
wmwv, line 374: Remove the following `#if err:` line

PR #26, opened 2025-04-24
Ricksuggestions

ColeFMeldorf:
In this PR we:
1. [x] Explicitly calculate magnitudes in the correct way that Lauren showed me  (now in calc_mags_and_err)
2. [x] Now save those magnitudes in the output lightcurve
3. [x] Calculate both the true and realized magnitude from the truth table (w/ and w/o the Poisson noise that Troxel told us is a little messed up.)

### DIFF
--- a/AllASPFuncs.py
+++ b/AllASPFuncs.py
@@ -773,12 +773,14 @@ def getWeights(cutout_wcs_list, size, snra, sndec, error=None,
 
         wgt = np.ones(size**2)
         wgt = 5*np.exp(-dist**2/gaussian_std)
-        wgt[np.where(dist > 4)] = 0
+        wgt[np.where(dist > 4)] = 0 # Correction here for flux missed ??? TODO
 
         if not isinstance(error, np.ndarray):
             error = np.ones_like(wgt)
-        wgt /= error[i*size**2:(i+1)*size**2]
-        wgt = wgt / np.sum(wgt)
+
+        wgt /= error[i*size**2:(i+1)*size**2] # Define an inv variance TODO
+        wgt = wgt / np.sum(wgt) # Normalize outside out of the loop TODO
+        # What fraction of the flux is contained in the PSF? TODO
 
         wgt_matrix.append(wgt)
     return wgt_matrix
@@ -1192,11 +1194,7 @@ def build_lightcurve(ID, exposures, sn_path, confusion_metric, flux,
     band (str): the bandpass of the images used
 
     Returns:
-    lc: a pandas dataframe containing the lightcurve data
-    Notes:
-    1.) This will soon be ECSV format instead
-    2.) Soon I will turn many of these inputs into environment variable and they
-    should be deleted from function arguments and docstring.
+    lc: a QTable containing the lightcurve data
     '''
 
     detections = exposures[np.where(exposures['DETECTED'])]
@@ -1204,6 +1202,11 @@ def build_lightcurve(ID, exposures, sn_path, confusion_metric, flux,
     df = open_parq(parq_file, path = sn_path, obj_type = object_type)
 
     mags, magerr = calc_mags_and_err(flux, sigma_flux, band)
+    zeros = np.zeros_like(detections['realized flux'])
+    sim_realized_mags, _ = calc_mags_and_err(detections['realized flux'],
+                                             zeros, band)
+    sim_true_mags, _ = calc_mags_and_err(detections['true flux'],
+                                         zeros, band)
 
     if object_type == 'SN':
         meta_dict ={'confusion_metric': confusion_metric, \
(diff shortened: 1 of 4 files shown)


wmwv:
Please add description of what the requested changes are.

AllASPFuncs.py
```diff
@@ -1157,6 +1154,24 @@ def contourGrid(image, numlevels = 5, subsize = 4):
...
+            'R062': 161.025,
+            'Y106': 302.275,
+            'Z087': 101.7}
+
+    area_eff = roman.collecting_area
+    galsim_zp = roman.getBandpasses()[band].zeropoint
+    mags = -2.5*np.log10(flux) + 2.5*np.log10(exptime[band]*area_eff) \
+        + galsim_zp
+
+    magerr = 2.5 * sigma_flux / (flux * np.log(10))
```
wmwv, line 1171: I would have written this as
```
magerr = (2.5 / np.log(10)) * sigma_flux / flux
```
I find these easier to read.  The first term captures the coefficient different between 2.5 log_10 and ln, while the second is really the measurement.

AllASPFuncs.py
```diff
@@ -1174,17 +1189,20 @@ def build_lightcurve(ID, exposures, sn_path, confusion_metric, flux,
...
-    should be deleted from function arguments and docstring.
+    lc: a QTable containing the lightcurve data
     '''
 
     detections = exposures[np.where(exposures['DETECTED'])]
     parq_file = find_parq(ID, path = sn_path, obj_type = object_type)
     df = open_parq(parq_file, path = sn_path, obj_type = object_type)
 
+    mags, magerr = calc_mags_and_err(flux, sigma_flux, band)
+    zeros = np.zeros_like(detections['realized flux'])
```
wmwv, line 1200: I would write these lines as
```
sim_sigma_flux = 0
sim_realized_mag, _ = calc_mags_and_err(detections["realized flux"], sim_sigma_flux, band)
sim_true_mags, _ = calc_mags_and_err(detections["true flux"], sim_sigma_flux, band)
```

You can use a scalar `0` instead of constructing an array.

AllASPFuncs.py
```diff
@@ -1194,14 +1212,19 @@ def build_lightcurve(ID, exposures, sn_path, confusion_metric, flux,
...
-        meta_dict = {'ra': df[df['id'] == str(ID)]['ra'].values[0], \
-            'dec': df[df['id'] == str(ID)]['dec'].values[0]}
-
-    data_dict = {'MJD': detections['date'], 'true_flux':
-    detections['realized flux'],  'measured_flux': flux, 'flux_error':
-    sigma_flux}
-    units = {'MJD':u.d, 'true_flux': '',  'measured_flux': '',
-             'flux_error': ''}
+        meta_dict = {'ra': df[df['id'] == str(ID)]['ra'].values[0],
+                     'dec': df[df['id'] == str(ID)]['dec'].values[0]}
```
wmwv, line 1216: Why are some of these `df["host_ra"][df["id"] == ID]` and some `df[df["id"] == str(ID)]]["host_ra"]`?

In particular, if the second works, I would recommend just extracting that row for `this_one = df[df["id"] == str(ID)]` and then refer to `this_one["ra"].values[0]`, etc.  (Pick a better name than `this_one`).

The idea here is both improved readability, and translating typos into things that you will catch because the variable isn't defined and your linter will catch, instead of a repeated test that you might get wrong and triggers a problem at run time but your linter can't catch.

AllASPFuncs.py
```diff
@@ -1221,10 +1244,10 @@ def build_lightcurve_sim(supernova, flux, sigma_flux):
     2.) Soon I will turn many of these inputs into environment variable and they
     should be deleted from function arguments and docstring.
     '''
-    data_dict = {'MJD': np.arange(0, detim, 1), 'true_flux': supernova,
-          'measured_flux':flux , 'flux_error': sigma_flux}
+    data_dict = {'MJD': np.arange(0, detim, 1), 'flux':flux,
```
wmwv, line 1247: I would prefer defining the MJD for the sims on a different line and then using that variable in the `data_dict` construction.

```
sim_mjd = np.arange(0, detim, 1)
data_dict = {"MJD": sim_mjd, ...
```

Also, insert a space in between `"flux": flux"`

config.yaml
```diff
@@ -7,7 +7,7 @@ check_perfection: false
 deltafcn_profile: false
 do_rotation: true
 do_xshift: true
-fetch_SED: false
+fetch_SED: true
```
wmwv, line 10: Why are these config changes part of this PR?

ColeFMeldorf: Apologies didn't mean to add that. Changes restored.

PR #33, opened 2025-04-28
Add lager

ColeFMeldorf:
@rknop suggested skipping making a test for now but all print statements are now lager outputs instead!

### DIFF
--- a/ASP_tests.py
+++ b/ASP_tests.py
@@ -134,6 +134,9 @@ def test_savelightcurve():
     lc_file = os.path.join(output_path, 'test_test_test_lc.ecsv')
     assert os.path.exists(lc_file) == True
 
+
 def test_run_on_star():
     os.system('python RomanASP.py -s 40973149150 -b Y106 -t 1 -d 1')
 
+
+def test_Lager():
--- a/AllASPFuncs.py
+++ b/AllASPFuncs.py
@@ -22,7 +22,9 @@ import galsim
 import h5py
 from scipy.interpolate import RegularGridInterpolator
 
-from snappl.image import OpenUniverse2024FTISImage
+
+from snappl.image import OpenUniverse2024FITSImage
+from snappl.logger import Lager
 
 pd.options.mode.chained_assignment = None  # default='warn'
 warnings.simplefilter('ignore', category=AstropyWarning)
@@ -72,7 +74,8 @@ def local_grid(ra_center, dec_center, wcs, npoints, size = 25, spacing = 1.0, im
     else:
         spacing = 1.0
 
-    print('GRID SPACE', spacing)
+    Lager.debug(f'GRID SPACE {spacing}')
+
     x = np.arange(difference, subsize+difference, spacing)
     y = np.arange(difference, subsize+difference, spacing)
 
@@ -83,10 +86,8 @@ def local_grid(ra_center, dec_center, wcs, npoints, size = 25, spacing = 1.0, im
         imcopy[imcopy <= 0] = 1e-10
         bins = [-np.inf]
         if len(percentiles) == 0:
-            #percentiles = [25, 80, 90]
             percentiles = [45, 90]
         bins.extend(np.nanpercentile(np.log(imcopy[np.where(np.log10(imcopy)>-10)]), percentiles))
-        #print('added 90 percentile to local grid')
         bins.append(np.inf)
 
         a = np.digitize(np.log(np.copy(imcopy)),bins)
@@ -118,14 +119,12 @@ def local_grid(ra_center, dec_center, wcs, npoints, size = 25, spacing = 1.0, im
         yy = np.array(ys)
 
     elif image is not None and makecontourGrid:
-        print('USING CONTOUR GRID')
+        Lager.debug('USING CONTOUR GRID')
         xx, yy = contourGrid(image)
         xx = np.array(xx)
         yy = np.array(yy)
 
-
     else:
-
         xx, yy = np.meshgrid(x+1, y+1)
     '''
         subsize = 8 #Taking a smaller square inside the image to fit on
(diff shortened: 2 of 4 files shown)


ASP_tests.py
```diff
@@ -144,6 +144,7 @@ def test_run_on_star():
     os.system('python RomanASP.py -s 40973149150 -b Y106 -t 1 -d 1')
 
 
+
```
wmwv, line 147: Why the inserted extra line?

AllASPFuncs.py
```diff
@@ -22,6 +22,10 @@
 import h5py
 from scipy.interpolate import RegularGridInterpolator
 
+
```
wmwv, line 25: Why the inserted extra line?

AllASPFuncs.py
```diff
@@ -141,20 +141,18 @@ def local_grid(ra_center, dec_center, wcs, npoints, size = 25, spacing = 1.0, im
...
+    Lager.debug(f'Built a grid with {np.size(xx)} points')
 
     if type(wcs)==galsim.fitswcs.AstropyWCS:
-        result = wcs.toWorld(xx, yy, units = 'deg')
+        result = wcs.toWorld(xx, yy, units='deg')
         ra_grid = result[0]
         dec_grid = result[1]
     else:
-        print('swapped x and y here')
+        Lager.warn('swapped x and y here')
```
wmwv, line 155: I suppose not a question for this PR, but why is this necessary.  What are there different conventions?  I would have thought the x, y convention would have already been included in the WCS matrix.

AllASPFuncs.py
```diff
@@ -244,8 +242,7 @@ def construct_psf_background(ra, dec, wcs, x_loc, y_loc, stampsize, bpass,
 
     k = 0
 
-    #For now, we use a flat SED. This is not ideal, but it is a good starting point.
-    print('In construct psf bg using flat SED')
+    Lager.debug('In construct psf bg using flat SED')
```
wmwv, line 245: I don't understand what this string means.  Are you annotating that `construct_psf_background` function is running?

AllASPFuncs.py
```diff
@@ -496,16 +488,20 @@ def construct_psf_source(x, y, pointing, SCA, stampsize=25,  x_center = None, y_
...
 
     config_file = './temp_tds.yaml'
     util_ref = roman_utils(config_file=config_file, visit = pointing, sca=SCA)
 
     assert sed is not None, 'You must provide an SED for the source'
 
     if not photOps:
-        print('WARNING: NOT USING PHOTON OPS IN PSF SOURCE')
-        print('ARE YOU SURE YOU WANT TO DO THIS?')
+        Lager.warn('NOT USING PHOTON OPS IN PSF SOURCE')
```
wmwv, line 504: If a user sets an option, why is there a warning about it?

AllASPFuncs.py
```diff
@@ -605,28 +610,12 @@ def constructImages(exposures, ra, dec, size = 7, background = False, roman_path
...
             im -= image[1].header['SKY_MEAN']
-            #print('subtracting sky mean not my fit')
-            #print('subtracted a bg of', popt[1])
-            #print('compared to: ', image[1].header['SKY_MEAN'])
-            #print('--------------------')
         elif not background and truth == 'truth':
             im -= bg
-            #print('Subtracted a BG of', bg)
-
+            Lager.debug(f'Subtracted a BG of {bg}')
```
wmwv, line 618: spell out "BG" -> "background"

AllASPFuncs.py
```diff
@@ -684,16 +673,15 @@ def getPSF_Image(self,stamp_size,x=None,y=None, x_center = None, y_center= None,
         return psf.drawImage(self.bpass,image=stamp,wcs=wcs,method='no_pixel',center = galsim.PositionD(x_center, y_center),use_true_center = True)
 
     photon_ops = [self.getPSF(x,y,pupil_bin)] + self.photon_ops
-    print('Using 1e6 photons in getPSF_Image')
+    Lager.debug('Using 1e6 photons in getPSF_Image')
```
wmwv, line 676: Is 1e6 set in a variable?  Use that variable for this warning message.

AllASPFuncs.py
```diff
@@ -684,16 +673,15 @@ def getPSF_Image(self,stamp_size,x=None,y=None, x_center = None, y_center= None,
...
     result = point.drawImage(self.bpass,wcs=wcs, method='phot', photon_ops=photon_ops, rng=self.rng, \
         n_photons=int(1e6),maxN=int(1e6),poisson_flux=False, center = galsim.PositionD(x_center, y_center),use_true_center = True, image=stamp)
     return result
 
 def fetchImages(testnum, detim, ID, sn_path, band, size, fit_background, roman_path):
-    #global object_type
     if len(str(ID)) != 8:
         object_type = 'star'
-        print('WARNING: Fitting SMP on a star. You probably want the grid off.')
+
```
wmwv, line 684: Why the blank line?

AllASPFuncs.py
```diff
@@ -1116,8 +1102,9 @@ def get_SN_SED(SNID, date, sn_path):
     mjd = sed_table['mjd']
     bestindex = np.argmin(np.abs(np.array(mjd) - date))
     if np.min(np.abs(np.array(mjd) - date)) > 10:
-        print('WARNING: No SED data within 10 days of date. \n \
-            The closest SED is ' + str(np.min(np.abs(np.array(mjd) - date))) + ' days away.')
+        Lager.warn('WARNING: No SED data within 10 days of date. \n \
+            The closest SED is ' + str(np.min(np.abs(np.array(mjd) - date))) +
+                                       ' days away.')
```
wmwv, line 1107: As a general practice for having tests and error messages, use a variable to make sure you have the same quantity and the error message and the drift don't fall out of sync.

You use the quantities
`10` and
`np.min(np.abs(np.array(mjd) - date))`
in both the test and the error message.  I suggest storing each as a variable, and then use the same variable in the test and the error message.

wmwv (changes requested):

AllASPFuncs.py
ColeFMeldorf: You're right that it's not extremely necessary, cutting it.

AllASPFuncs.py
ColeFMeldorf: My logic was that, while I want to do this sometimes, it is very rare that you actually want to do this. Thus if it was accidentally on while doing a normal run, I'd want to know. I can delete it but I found it useful for me. It could be .debug instead?

AllASPFuncs.py
ColeFMeldorf: Yes there seem to be different conventions, unless I made a mistake. This type of WCS isn't typically used by the code, this branch of the else statement is never used. I am planning on fixing on all of the WCS convention strangeness when I implementnt the snappl image class if that's ok, as then it will all be standardized for once.

AllASPFuncs.py
wmwv: Totally fine.   Take the sentence above in your comment and put it as a comment in the code.

AllASPFuncs.py
wmwv: Yes, I was just flagging this for the future.

PR #36, opened 2025-04-28
Oops, wasn't actually fixed! Addendum to previous PR!

ColeFMeldorf:
In the regression test, the config file was not actually being written to before running the test, causing it to fail. This has now been fixed. Also, I made it so that it now checks if the flux error matches the saved lightcurve as well.
You may notice the confusion metric is different in each file. That's fine because the confusion_metric still has randomness in it and is not being used as a part of the test.

### DIFF
--- a/RomanASP.py
+++ b/RomanASP.py
@@ -48,8 +48,6 @@ Adapted from code by Pedro Bernardinelli
 '''
 
 
-
-
 def load_config(config_path):
     """Load parameters from a YAML configuration file."""
     with open(config_path, 'r') as file:
@@ -58,7 +56,6 @@ def load_config(config_path):
 
 
 def main():
-
     parser = argparse.ArgumentParser(description="Can overwrite config file")
 
     parser.add_argument('-b', '--band', type=str, required=True, help='filter')
@@ -117,8 +114,6 @@ def main():
     mismatch_seds = config['mismatch_seds']
     fetch_SED = config['fetch_SED']
     makecontourGrid = config['makecontourGrid']
-
-
 
     roman_bandpasses = galsim.roman.getBandpasses()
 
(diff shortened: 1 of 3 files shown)


wmwv (changes requested):
I agree with Copilot's suggestion

RomanASP.py
```diff
@@ -47,8 +47,7 @@
 
 '''
 
-config_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
-                           'config.yaml')
+
```
wmwv, line 50: Remove extra blank lines.  Should only be two before the first `def`

RomanASP.py
```diff
@@ -75,6 +74,21 @@ def main():
...
+    SNID = args.SNID
+    testnum = args.testnum
+    detim = args.detim
+    output_path = args.output_path
+    if args.config is not None:
+        config_path = args.config
+    else:
+        config_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                           'config.yaml')
+
```
wmwv, line 91: Someday it would be nice to separate out parsing of arguments, loading of configs, and then running the main function.

RomanASP.py
```diff
@@ -104,12 +118,7 @@ def main():
     fetch_SED = config['fetch_SED']
     makecontourGrid = config['makecontourGrid']
 
-    args = parser.parse_args()
-    band = args.band
-    SNID = args.SNID
-    testnum = args.testnum
-    detim = args.detim
-    output_path = args.output_path
+
```
wmwv, line 121: Consider a habit change so that when you delete things you make sure that the number of blank lines left behind is reasonable.  As a starter, delete the entire section rather than replacing it with a blank line.

wmwv (approved):
Check extra blank lines.  The rest is editorial for future issues.

PR #37, opened 2025-04-28
Linalg

ColeFMeldorf:
All the linear algebra array reshaping is now done in one function, prep_data_for_fit.


wmwv:
The changes I see here include the ones for the logging.  Can you prepare a cleaner PR/branch that just shows the changes with respect to the nominal content of this branch?

PR #39, opened 2025-04-29
Linalg cherrypick

ColeFMeldorf:
This contains the same commits as #37 but with the intermediate changes from #33 removed for a cleaner history.

NB: There are a few places where Lager commands are modified still. This is because I realized that .warn() is not the right command, it should be .warning().


AllASPFuncs.py
```diff
@@ -620,13 +620,23 @@ def constructImages(exposures, ra, dec, size = 7, background = False, roman_path
             im -= bg
             Lager.debug(f'Subtracted a background level of {bg}')
 
-        m.append(im.flatten())
-        err.append(err_cutout.flatten())
-        mask.append(np.zeros(size*size))
-        #w = (zero**2)*err_cutout.flatten()
+        #m.append(im.flatten())
```
wmwv, line 623: What's happening in this section?  Why are some lines commented out?  Is this something you're still deciding about?  Why does it say "for not"?

If you're not sure if this is the path you want, then you can write out each as a different case right here and have an option flag to trigger each behavior.

Otherwise, I suggest remove the commented-out lines.

If you're keeping the lines because you might want them someday, then make them accessible as something you can call with an option to active.  Commented-out means you have to edit the code to run that, which means you can't write a test that exercises the section against a specific version.

AllASPFuncs.py
```diff
@@ -1292,3 +1307,34 @@ def banner(text):
     length = len(text) + 8
     message = "\n" + "#" * length +'\n'+'#   ' + text + '   # \n'+ "#" * length
     Lager.debug(message)
+
+
+def prep_data_for_fit(images, err, sn_matrix, wgt_matrix):
+    '''
+    This function takes the data from the images and puts it into the form such
+    that we can analytically solve for the best fit using linear algebra.
```
wmwv, line 1315: Write out inputs and outputs.  Because the purpose of this function is explicitly to transfer a set of inputs into the same content but in a different structure, clearly explaining the expected form of the input and output is a key part of understanding the function.

AllASPFuncs.py
```diff
@@ -1292,3 +1307,34 @@ def banner(text):
...
+    '''
+    size = int(np.sqrt((images[0].size)))
```
wmwv, line 1317: You only use size_sq below, so I would just calculate size_sq.  Or even perhaps num_pixels.

I bring this up both because 
a. it's simpler
b. but also you're relying here that a float sqrt operation followed by an int squaring operation will be reversible, but that's not guaranteed to work.  This happens to work in Python and numpy if N is a square, but is not guaranteed.
c. The reason that this caught my eye was that I often write code to do something simple like make a tiling of plots, where N may not be a square.  There I want `size = int(sqrt(N)) + 1` to make sure I have a square that includes all of the N.

a = int(sqrt(a*a))

AllASPFuncs.py
```diff
@@ -1292,3 +1307,34 @@ def banner(text):
...
+
+    for i in range(det_num):
+        psf_zeros[
+            (tot_num - det_num + i) * size * size:
+            (tot_num - det_num + i + 1) * size * size,
+            (tot_num - det_num) + i] = sn_matrix[i]
+    sn_matrix = psf_zeros
+    sn_matrix = np.array(sn_matrix)
+    sn_matrix = np.vstack(sn_matrix)
+
```
wmwv, line 1335: Remove extra space.

AllASPFuncs.py
```diff
@@ -1292,3 +1307,34 @@ def banner(text):
...
+    err = np.concatenate([arr.flatten() for arr in err])
+
+    psf_zeros = np.zeros((np.size(images), tot_num))
+
+    for i in range(det_num):
+        psf_zeros[
+            (tot_num - det_num + i) * size * size:
+            (tot_num - det_num + i + 1) * size * size,
+            (tot_num - det_num) + i] = sn_matrix[i]
+    sn_matrix = psf_zeros
```
wmwv, line 1332: You can replace these three lines with just

```
sn_matrix = np.vstack(psf_zeros)
```

psf_zeros is definitely an numpy array, you just created it above.  Also, for cases where you're not sure if something is an array or not, use `np.as_array` instead of `np.array`.  The former, `np.as_array` will only create a new object if necessary, where as `np.array` will always create a new array.  For small arrays, this doesn't matter, but you're doing linear-algebra where you should be starting to think about having large arrays and some of the performance implications.

AllASPFuncs.py
```diff
@@ -1292,3 +1307,34 @@ def banner(text):
...
+    tot_num = len(images)
+    det_num = len(sn_matrix)
+
+    # Flatten into 1D arrays
+    images = np.concatenate([arr.flatten() for arr in images])
+    err = np.concatenate([arr.flatten() for arr in err])
+
+    psf_zeros = np.zeros((np.size(images), tot_num))
+
+    for i in range(det_num):
```
wmwv, line 1327: Define `tot_num - det_num + i` as a variable, and then use that variable in the indexing.
This is better because you're not re-typing an operation (I had to double-check that `(tot_num - det_num + i)` and `(tot_num - det_num) + i` were the same.) and because it will actually make the indexing shorter and easier to read.

AllASPFuncs.py
```diff
@@ -1292,3 +1307,34 @@ def banner(text):
...
+    size = int(np.sqrt((images[0].size)))
+    tot_num = len(images)
+    det_num = len(sn_matrix)
+
+    # Flatten into 1D arrays
+    images = np.concatenate([arr.flatten() for arr in images])
+    err = np.concatenate([arr.flatten() for arr in err])
+
+    psf_zeros = np.zeros((np.size(images), tot_num))
+
```
wmwv, line 1326: Explain what this is doing with the psf_zeros and the selecting filling in the for loop.

RomanASP.py
```diff
@@ -19,10 +19,14 @@
 import scipy.sparse as sp
 from numpy.linalg import LinAlgError
 import galsim
+import AllASPFuncs
+import importlib
+importlib.reload(AllASPFuncs)
```
wmwv, line 24: These importlib lines shouldn't be necessary.  If they are, something is wrong.

simulation.py
```diff
@@ -159,10 +159,14 @@ def simulate_images(testnum, detim, ra, dec, do_xshift, do_rotation, supernova,
                 sn_storage.append(supernova_image)
 
         cutout_wcs_list.append(cutoutgalwcs)
-        imagelist.append(a.flatten())
+        # No longer flattening
```
wmwv, line 162: Same comment about either leaving commented-out code exposed through a code option or removing.

simulation.py
```diff
@@ -227,7 +231,7 @@ def simulate_supernova(snx, sny, stamp, flux, sed, band, sim_psf,
 
     # Code below copied from galsim largely
     if not source_phot_ops:
-        result = sim_psf.drawImage(roman_bandpasses[band], image=stamp,
+        result = profile.drawImage(roman_bandpasses[band], image=stamp,
```
wmwv, line 234: Why is this change in this PR?

wmwv (changes requested):
Assorted minor comments to address.  Should be quick.

simulation.py
ColeFMeldorf: Good point, this should go in a different one. Removed.

ColeFMeldorf:
All requests addressed!

wmwv:
I translate this as "This PR is ready for another review".

ColeFMeldorf:
Correct! Ideally it shouldn't need much looking at because I just implemented your changes but yes I meant it is ready for another review!

AllASPFuncs.py
```diff
@@ -1292,3 +1295,60 @@ def banner(text):
...
+        sn_index = tot_num - det_num + i # We only want to edit SN columns.
+        psf_zeros[
+            (sn_index) * size_sq:  # Fill in rows s^2 * image number...
+            (sn_index + 1) * size_sq, #... to s^2 * (image number + 1) ...
+            sn_index] = sn_matrix[i] # ...in the correct column.
+    sn_matrix = np.vstack(psf_zeros)
+    wgt_matrix = np.array(wgt_matrix)
+    wgt_matrix = np.hstack(wgt_matrix)
+
+    print(np.shape(images), np.shape(err), np.shape(sn_matrix), np.shape(wgt_matrix))
```
wmwv, line 1353: Lager?

RomanASP.py
```diff
@@ -19,10 +19,12 @@
 import scipy.sparse as sp
 from numpy.linalg import LinAlgError
 import galsim
+import AllASPFuncs
```
wmwv, line 22: Why both `import AllASPFuncts` when you have `from AllASPFuncs ...` just below?

wmwv (changes requested):
two minor comments.

PR #42, opened 2025-04-29
Fixed error bug

ColeFMeldorf:
Addressed both tasks in the description.

### DIFF
--- a/AllASPFuncs.py
+++ b/AllASPFuncs.py
@@ -759,6 +759,14 @@ def getWeights(cutout_wcs_list, size, snra, sndec, error=None,
         wgt = np.ones(size**2)
         wgt = 5*np.exp(-dist**2/gaussian_std)
 
+        # Here, we throw out pixels that are more than 4 pixels away from the
+        # SN. The reason we do this is because by choosing an image size one
+        # has set a square top hat function centered on the SN. When that image
+        # is rotated pixels in the corners leave the image, and new pixels
+        # enter. By making a circular cutout, we minimize this problem. Of
+        # course this is not a perfect solution, because the pixellation of the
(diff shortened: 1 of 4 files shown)


AllASPFuncs.py
```diff
@@ -501,7 +501,7 @@ def construct_psf_source(x, y, pointing, SCA, stampsize=25,  x_center = None, y_
         # While I want to do this sometimes, it is very rare that you actually
         # want to do this. Thus if it was accidentally on while doing a normal
         # run, I'd want to know.
-        Lager.warn('NOT USING PHOTON OPS IN PSF SOURCE')
+        Lager.warning('NOT USING PHOTON OPS IN PSF SOURCE')
```
wmwv, line 504: I feel like the SED PR also includes this.

AllASPFuncs.py
```diff
@@ -741,30 +741,31 @@ def get_object_info(ID, parq, band, snpath, roman_path, obj_type):
...
-        snx, sny = wcs.toImage(snra, sndec, units = 'deg')
+        snx, sny = wcs.toImage(snra, sndec, units='deg')
         dist = np.sqrt((xx - snx + 1)**2 + (yy - sny + 1)**2)
 
         wgt = np.ones(size**2)
-
-
         wgt = 5*np.exp(-dist**2/gaussian_std)
+
         wgt[np.where(dist > 4)] = 0
```
wmwv, line 762: Explain why the units of dist (wr.t. the bare value `4`) and why it's OK to throw out outliers and whether that affects the calculation.

wmwv (approved):
LGTM

Think through your git branching workflow to check why you have some repeated changes in two different PRs.

PR #44, opened 2025-04-29
Get sed improvement

ColeFMeldorf:
See description, all tasks completed


AllASPFuncs.py
```diff
@@ -1255,3 +1264,18 @@ def banner(text):
     length = len(text) + 8
     message = "\n" + "#" * length +'\n'+'#   ' + text + '   # \n'+ "#" * length
     Lager.debug(message)
+
+def get_SED_list(ID, exposures, fetch_SED, object_type, sn_path):
+    sedlist = []
+    for date in exposures['date'][exposures['DETECTED']]:
```
wmwv, line 1270: I have a naming/design question:   Why doesn't `get_SED_list` doesn't return of list of things returned by `get_SED`.

Think a bit about the naming of

get_star_SED
get_SN_SED
get_SED
get_SED_list

to see if when you try to explain it to someone (even if that's just in writing up the documentation) you feel it makes sense.

AllASPFuncs.py
```diff
@@ -1255,3 +1264,18 @@ def banner(text):
     length = len(text) + 8
     message = "\n" + "#" * length +'\n'+'#   ' + text + '   # \n'+ "#" * length
     Lager.debug(message)
+
+def get_SED_list(ID, exposures, fetch_SED, object_type, sn_path):
```
wmwv, line 1268: Add doc string explaining what function does, what it takes as input, and what it produces as output.

RomanASP.py
```diff
@@ -279,7 +265,8 @@ def main():
             snx, sny = cutout_wcs_list[0].toImage(snra, sndec, units='deg')
             pointing, SCA = exposures['Pointing'][0], exposures['SCA'][0]
             array = construct_psf_source(x, y, pointing, SCA, stampsize=size,
-                                         x_center=snx, y_center=sny, sed=sed)
+                                         x_center=snx, y_center=sny,
+                                         sed=sedlist[0])
```
wmwv, line 269: Maybe add a comment to this section here that calculates the confusion metric that explains why (and that) you only use the first entry.

RomanASP.py
```diff
@@ -347,12 +334,8 @@ def main():
                     else:
                         pointing = 662
                         SCA = 11
-                    if fetch_SED:
 
-                        Lager.debug(f'Using SED #{str(i - (testnum - detim))}')
-                        sed = sedlist[i - (testnum - detim)]
-                    else:
-                        Lager.debug('Using default SED')
+                    sed = sedlist[i - (testnum - detim)]
```
wmwv, line 338: Explain why this is the index you want.

wmwv (changes requested):
Write out some doc strings for get_SED_list and the other get_SED and get_*_SED functions and while doing that see if you can standardize on a naming scheme that makes sense.  To me, I think `get_SED_list` should construct a list by calling `get_SED`, but that's not actually what this does, which I find confusing.

AllASPFuncs.py
ColeFMeldorf: I'm a little confused because get_SED_list does return a list of SEDs? or do you mean it sounds too much like the previous function and should instead be get_list_of_SEDs, or something?

wmwv: `get_SED` returns a tuple of arrays, one of wavelength, one of fluxdensity.
`get_SED_list` returns a list of galsim.SED objects.

This is confusing.

RomanASP.py
```diff
@@ -247,6 +232,7 @@ def main():
             dec_grid = np.array([])
 
         # Get the weights
+        Lager.warning('SURPRESSING ERROR UNTIL FIX PUSHED TO MAIN')
```
wmwv, line 235: "SURPRESSING ERROR CALCULATION UNTIL..."?

RomanASP.py
```diff
@@ -347,12 +334,14 @@ def main():
...
-                        sed = sedlist[i - (testnum - detim)]
-                    else:
-                        Lager.debug('Using default SED')
+                    # sedlist is the length of the number of supernova
+                    # detection images. Therefore, when we iterate onto the
+                    # first supernova image, we want to be on the first element
+                    # of sedlist. Therefore, we subtract by the number of
+                    # predetection images: testnum - detim.
+                    sn_index = i - (testnum - detim)
+                    Lager.debug(f'Using SED #{str(sn_index)}')
```
wmwv, line 343: If you're using a format string anyway then write this as a format rather than a `str` call

Lager.debug(f"Using SED #{sn_index:s}")

ColeFMeldorf: Didn't know about this, thanks!

AllASPFuncs.py
ColeFMeldorf: Got it, new name is now get_all_galsim_SEDs

wmwv: My suggested function names are `get_galsim_SED` and `get_galsim_SED_list`.

Write a function that takes part of the logic in the for loop of the current `get_SED_list` and make that `get_galsim_SED`.  And then `get_galsim_SED_list` iterates over the list calling `get_galsim_SED` each time.

ColeFMeldorf: Love it, done!

AllASPFuncs.py
```diff
@@ -1069,22 +1076,46 @@ def slice_plot(fileroot):
...
 
+    Internal Variables:
+    lam: the wavelength of the SED in Angstrom
+    flambda: the flux of the SED units in erg/s/cm^2/Angstrom
 
-
-def get_SED(SNID, date, sn_path, obj_type = 'SN'):
-    #Is this an ok way to do this?
+    Returns:
+    sed: the SED of the object in galsim's format
```
wmwv, line 1095: You could also say something like Returns a `galsim.SED` object

AllASPFuncs.py
```diff
@@ -1069,22 +1076,46 @@ def slice_plot(fileroot):
...
 
 
 def get_star_SED(SNID, sn_path):
+    '''
+    Return the appropriate SED for the star.
+    Inputs:
+    SNID: the ID of the object
+    sn_path: the path to the supernova data
+
+    Returns:
```
wmwv, line 1115: Maybe clarify that these are expected to be numpy.arrays of floats.

AllASPFuncs.py
```diff
@@ -1264,3 +1307,36 @@ def banner(text):
     length = len(text) + 8
     message = "\n" + "#" * length +'\n'+'#   ' + text + '   # \n'+ "#" * length
     Lager.debug(message)
+
+def get_galsim_SED_list(ID, exposures, fetch_SED, object_type, sn_path):
+    sedlist = []
+    '''
+    Return the appropriate SED for the object for each observation.
+    If you are getting truth SEDs, This function calls get_SED on each exposure
```
wmwv, line 1315: ", This" -> ", this"

AllASPFuncs.py
wmwv: Where?  I don't see this change.  I still see a for loop over two options within the loop instead of over one function.

If you really want to keep this this way, I would suggest that you at least make the if/else higher than the for loop level.  The if will be true/false for the entire set.  It's not a row-by-row thing.

wmwv (changes requested):
Once last nudge about having a function that returns an SED for one SN+date.  And then iterating over that function.  You still have code in the for loop that's difficult to test on its own.

ColeFMeldorf:
This is also ready for re-review!

wmwv (approved):
LGTM

PR #48, opened 2025-04-30
Snappl updates

ColeFMeldorf:
In this PR I include the first batch of refactoring to use the snappl image class. '

1. [x] refactor constructImages to use snappl

Since this will be a massive refactor, I chose to do this in batches so that your reviews take a few minutes and not an hour.


wmwv (changes requested):
If you're going to start a re-factor, than I strongly suggest you merge other open branches into main, pause, and then start the re-factor.  This particular issue branches off of 549e7c3 and not current origin/main.

ColeFMeldorf:
Ok! All other branches closed. Let me know what you think!

wmwv (approved):
LGTM

PR #52, opened 2025-05-06
Rickbugfixes

ColeFMeldorf:
This PR accomplishes a few small goals as given to me by Rick, so this is a grab bag of small little changes:
1. [x] Rick pointed out that weights should be normalized outside of the loop in which they were created, this caused me to realize they don't need to be normalize at all. This seems to make the error be the right order of magnitude now!
2. [x] While I don't believe this will matter, I am now outputting how much of the PSF volume is lost by my 4 pixel distance cut.
3. [x] You can now pass a time range to the code, say -20 20, and it will perform SMP on that SN from 20 days before to 20 days after peak brightness.
4. [x] plot_lc now plots the flux error too, and fixed it reading the wrong file.
5. [x] Now that error is calculated correctly, I needed to update the test lightcurve for the regression test.

### DIFF
--- a/AllASPFuncs.py
+++ b/AllASPFuncs.py
@@ -379,15 +379,11 @@ def findAllExposures(snid, ra, dec, peak, start, end, band, maxbg=24,
         bg = bg.iloc[:maxbg]
     bg['DETECTED'] = False
 
-    #combine these two dataframes
+    # combine these two dataframes
     all_images = pd.concat([det, bg])
     all_images['zeropoint'] = np.nan
 
-    Lager.debug(all_images['offpeak_time'])
-    Lager.debug(lc_start)
-    Lager.debug(lc_end)
-
-    #Now we need to loop through the images and get the information we need
+    # Now we need to loop through the images and get the information we need
     zpts = []
     true_mags = []
     true_fluxes = []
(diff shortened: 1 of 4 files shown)


AllASPFuncs.py
```diff
@@ -354,27 +353,40 @@ def findAllExposures(snid, ra,dec,peak,start,end,band, maxbg = 24, maxdet = 24,
...
+                    (bg['offpeak_time'] <= lc_end)]
+    if isinstance(maxbg, int):
+        bg = bg.iloc[:maxbg]
     bg['DETECTED'] = False
 
     #combine these two dataframes
     all_images = pd.concat([det, bg])
     all_images['zeropoint'] = np.nan
 
+    Lager.debug(all_images['offpeak_time'])
```
wmwv, line 386: Document what the thing being printed is in these debug messages.

AllASPFuncs.py
```diff
@@ -827,50 +845,47 @@ def makeGrid(adaptive_grid, images, size, ra, dec, cutout_wcs_list,
 
 def plot_lc(fileroot):
 
-    fluxdata = pd.read_csv('./results/lightcurves/'+str(fileroot)+'_lc.csv')
-    supernova = fluxdata['true_flux']
-    measured_flux = fluxdata['measured_flux']
+    fluxdata = pd.read_csv('./results/lightcurves/'+str(fileroot)+'_lc.ecsv',
```
wmwv, line 848: For a future issue, make this `./results/lightcurves/` more flexible and something that can be passed in more easily.  For now I might even suggest making the argument to `plot_lc` the full path name.

AllASPFuncs.py
```diff
@@ -827,50 +845,47 @@ def makeGrid(adaptive_grid, images, size, ra, dec, cutout_wcs_list,
...
     bias = np.round(bias, 3)
-    scatter = np.std(-2.5*np.log10(measured_flux)+2.5*np.log10(np.array(supernova)))
+    scatter = np.std(mag - truth_mags)
     scatter *= 1000
     scatter = np.round(scatter, 3)
     props = dict(boxstyle='round', facecolor='wheat', alpha=0.8)
     textstr = 'Overall Bias: ' + str(bias) + ' mmag \n' + \
         'Overall Scatter: ' + str(scatter) + ' mmag'
-    plt.text(np.percentile(dates,60), 14 - 2.5*np.log10(np.mean(supernova)), textstr,  fontsize=14,
+    plt.text(np.percentile(dates,60), np.mean(truth_mags), textstr,  fontsize=14,
```
wmwv, line 875: The magnitude of the mean of the flux and the mean of the magnitudes are the not the same thing.  Which do you want?  It's not obvious which you want.

AllASPFuncs.py
```diff
@@ -827,50 +845,47 @@ def makeGrid(adaptive_grid, images, size, ra, dec, cutout_wcs_list,
...
-    if flux_mode:
-        plt.scatter(dates, X[-detim:] - supernova, color = 'k')
-        for i,dr in enumerate(zip(dates, X[-detim:] - supernova)):
-            d,r = dr
-            plt.text(d+1,r,i+testnum-detim, fontsize = 8)
-    else:
-        plt.scatter(dates, -2.5*np.log10(measured_flux)+2.5*np.log10(supernova), color = 'k')
-        plt.axhline(0, ls = '--', color = 'k')
-        plt.ylabel('Mag Residuals (Model - Truth)')
+
```
wmwv, line 881: define `residual = mag - truth_mags` and use that variable in the error bar and the min, max calculations below.

RomanASP.py
```diff
@@ -140,14 +156,18 @@ def main():
     airy = galsim.ChromaticOpticalPSF(lam, diam=2.36,
                                       aberrations=aberrations)
 
+    Lager.warning('manually removing weird simulation leftovers for now')
```
wmwv, line 159: I don't understand what this would mean if I were reading a log file and say this line.  This seems like more of a comment about the code rather than a log message for the running.  What is this warning me of?

RomanASP.py
```diff
@@ -371,6 +409,13 @@ def main():
         images, err, sn_matrix, wgt_matrix =\
             prep_data_for_fit(images, err, sn_matrix, wgt_matrix)
 
+        # Estimate amount of the PSF cut out by setting a distance cap
```
wmwv, line 412: Why is this just an estimate.  How could it be made more accurate or precise?

wmwv (changes requested):
Add a test for the fix you made to plot_lc.

Assorted minor comments.

RomanASP.py
ColeFMeldorf: Great point, it is exact!

AllASPFuncs.py
ColeFMeldorf: I meant mean of truth mags!

PR #53, opened 2025-05-06
Moving Simulation stuff out of the main algorithm. 

ColeFMeldorf:
In this PR we:

### DIFF
--- a/ASP_tests.py
+++ b/ASP_tests.py
@@ -63,19 +63,35 @@ def test_findAllExposures():
 
 
 def test_simulate_images():
-    band = 'F184'
+
+    config_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                               'config.yaml')
+    config = yaml.safe_load(open(config_path))
+    config['band'] = 'F184'
+    config['do_xshift'] = True
+    config['do_rotation'] = True
+    config['size'] = 11
+    config['noise'] = 0
+    config['use_roman'] = False
+    config['use_real_images'] = False
+    config['fetch_SED'] = False
+    config['roman_path'] = roman_path
+    config['bg_gal_flux'] = 9e5
+    config['deltafcn_profile'] = False
+    config['sim_ra'] = 7.541534306163982
+    config['sim_dec'] = -44.219205940734625
+    config['base_pointing'] = 662
+    config['base_sca'] = 11
+
     lam = 1293  # nm
     airy = \
         galsim.ChromaticOpticalPSF(lam, diam=2.36, aberrations=galsim.roman.
-                                   getPSF(1, band, pupil_bin=1).aberrations)
-    images, im_wcs_list, cutout_wcs_list, psf_storage, sn_storage, sim_lc = \
-        simulate_images(testnum=10, detim=5, ra=7.541534306163982,
-                        dec=-44.219205940734625, do_xshift=True,
-                        do_rotation=True, sim_lc=[10, 100, 1000,
-                                                     10**4, 10**5],
-                        noise=0, use_roman=False, band='F184',
-                        deltafcn_profile=False, roman_path=roman_path, size=11,
-                        input_psf=airy, bg_gal_flux=9e5)
+                                   getPSF(1, config['band'],
+                                          pupil_bin=1).aberrations)
+    images, im_wcs_list, cutout_wcs_list, sim_lc, util_ref = \
+        simulate_images(testnum=10, detim=5, config=config,
+                        sim_lc=[10, 100, 1000, 10**4, 10**5],
+                        input_psf=airy)
 
     compare_images = np.load('tests/testdata/images.npy')
     assert compare_images.all() == np.asarray(images).all()
--- a/ASP_tests.py
+++ b/ASP_tests.py
@@ -68,10 +68,10 @@ def test_simulate_images():
     airy = \
         galsim.ChromaticOpticalPSF(lam, diam=2.36, aberrations=galsim.roman.
                                    getPSF(1, band, pupil_bin=1).aberrations)
-    images, im_wcs_list, cutout_wcs_list, psf_storage, sn_storage = \
+    images, im_wcs_list, cutout_wcs_list, psf_storage, sn_storage, sim_lc = \
         simulate_images(testnum=10, detim=5, ra=7.541534306163982,
                         dec=-44.219205940734625, do_xshift=True,
-                        do_rotation=True, supernova=[10, 100, 1000,
+                        do_rotation=True, sim_lc=[10, 100, 1000,
                                                      10**4, 10**5],
                         noise=0, use_roman=False, band='F184',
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -74,6 +75,9 @@ For testing the algorithm, it is often beneficial to simulate our own galaxy and
 | turn_grid_off          | bool   | If true, don't generate a background model at all. Useful for testing just the PSF photometry of the SN if bg_gal_flux is set to zero. |
 | single_grid_point      | bool   | See below.                                                                                                                           |
 | deltafcn_profile       | bool   | If true, the galaxy is no longer a realistic galaxy profile and instead a Dirac delta function. Combined with single_grid_point, it is hypothetically possible for the algorithm to perfectly recover the background by fitting a Dirac delta to a Dirac delta at the exact same location. TODO: explain this better. |
+|sim_ra, sim_dec         | float  | RA and DEC for simulated SN in degrees. |
+|base_pointing, base_sca | int    | Pointing and SCA for base roman image to use for simulation. For instance, this image is used to set the
```
wmwv, line 79: Roman

README.md
```diff
@@ -57,10 +57,11 @@ To actually have the code serve your specific needs, you can modify the yaml fil
 | fetch_SED              | bool   | If true, get the SN SED from the OpenUniverse parquet files. If false, use a flat SED. May not be perfectly functional yet. TODO: see if this is improvable. |
 | make_initial_guess     | bool   | If true, the algorithm uses an average of the pixel values at each model point to set an initial guess for each model point. Slight improvement in certain cases but not pivotal. |
 | source_phot_ops        | bool   | If true, use photon shooting to generate the PSF for fitting the SN. This seemingly needs to be true for a quality fit.     |
+| flux_initial_guess | float | For making the initial guess, what flux should the SN be set to? This makes practically zero difference to the results, there just needs to be a choice made. |
```
wmwv, line 60: Can you provide better guidance to the user?
"This makes practically zero difference to the results, there just needs to be a choice made." doesn't really help.  Should this just be `1`?  Does it ever matter if it's not `1`?

RomanASP.py
```diff
@@ -251,14 +232,12 @@ def main():
         # Otherwise, initializing the model guess does not make sense.
         # TODO: Are testnum and detim both ints? Then compare for equality.
         if make_initial_guess and testnum - detim != 0:
```
wmwv, line 234: What is `supernova` vs. `detim` and why is it changed here?

simulation.py
```diff
@@ -54,6 +50,21 @@ def simulate_images(testnum, detim, ra, dec, do_xshift, do_rotation, supernova,
     im_wcs_list: a list of the wcs objects for each full SCA image
     cutout_wcs_list: a list of the wcs objects for each cutout image
     '''
+
```
wmwv, line 53: If all of these are required arguments, my preference would be to just list them and make it obvious in the function signature that they are all required and then separately documented in the function docstring.

You use the same number of code lines, because all of the lines below to pull things out of the `config` object would instead just be part of the `def ...`

simulation.py
```diff
@@ -62,6 +73,15 @@ def simulate_images(testnum, detim, ra, dec, do_xshift, do_rotation, supernova,
     galra = ra + 1.5e-5
     galdec = dec + 1.5e-5
 
+    if sim_lc is None:
```
wmwv, line 76: Why?  Document.  The continued use of `detim`, which I don't understand, is continuing to trip me up every time I try to read and understand a code block.

ASP_tests.py
```diff
@@ -63,19 +63,35 @@ def test_findAllExposures():
...
-                        dec=-44.219205940734625, do_xshift=True,
-                        do_rotation=True, supernova=[10, 100, 1000,
-                                                     10**4, 10**5],
-                        noise=0, use_roman=False, band='F184',
-                        deltafcn_profile=False, roman_path=roman_path, size=11,
-                        input_psf=airy, bg_gal_flux=9e5)
+                                   getPSF(1, config['band'],
+                                          pupil_bin=1).aberrations)
+    images, im_wcs_list, cutout_wcs_list, sim_lc, util_ref = \
+        simulate_images(testnum=10, detim=5, config=config,
```
wmwv, line 92: See below about being explicitly about what it is  that `simulate_images` takes as arguments.

ASP_tests.py
```diff
@@ -63,19 +63,35 @@ def test_findAllExposures():
 
 
 def test_simulate_images():
-    band = 'F184'
+
+    config_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                               'config.yaml')
+    config = yaml.safe_load(open(config_path))
```
wmwv, line 69: What quantities do we need from the `config.yaml`?

wmwv (changes requested):
I don't like passing config objects around as a standard way to call a function.  It makes it much harder to understand what a function requires.  It's how you end up with dependencies on `tds.yaml`, where you only care about 5 out of 100 lines, but you don't know which ones.

It's okay to pass a config option around through a pipeline and passing through arguments you don't need to know about.  But if a function is going to directly use quantities, they should be part of the function signature.  That makes them much easier to understand that they're needed, what reasonable defaults might be, and eventually to type check them.

RomanASP.py
ColeFMeldorf: Stupidly, I was using the same variable, supernova, in two different ways depending on whether you were simulating or using OpenUniverse images. I have now separated out all the simulation stuff into simulation.py, and hence supernova had to be replaced with a variable that actually is used in the non simulation case. detim is the number of detection images, which I'll change to num_detected_images in the next refactor. But for now, this is making sure we don't try to run the initial guess algortihm on the case where there are no background images.

simulation.py
ColeFMeldorf: fixed!

ASP_tests.py
```diff
@@ -188,7 +188,7 @@ def test_regression():
             as temp_config:
         yaml.dump(config, temp_config)
         temp_config_path = temp_config.name
-    output = os.system(f'python RomanASP.py -s 40120913 -b Y106 -t 2 -d 1 -o \
+    output = os.system(f'python RomanASP.py -s 40120913 -f Y106 -t 2 -d 1 -o \
```
wmwv, line 191: Is there a documentation or logging change that would have helped you catch this earlier?

wmwv (changes requested):
Lager rather than print

Is there a documentation or arg-checking or tests change that would have helped you catch the bug more easily?

ASP_tests.py
ColeFMeldorf: This was not the bug fix that I commented on being the worst ever :)) However, rick did suggest that I not bother with the single letter arguments because it does get confusing. Therefore I plan to switch to whole word arguments only, but that's outside the scope of this PR.

PR #54, opened 2025-05-11
Onefunc and varnames

ColeFMeldorf:
1. [x] Package the giant for loop that runs my code into one function that runs a single supernova
2. [x] This should return data that can then be saved
3. [x] Check that this doesn't break any of the already written tests.
4. [x] To make it easier to review, I am now changing totnum --> num_total_imgs and detnum ---> num_detect_imgs.


OneObjFunc.py
```diff
@@ -0,0 +1,327 @@
...
+from simulation import simulate_images
+import yaml
+import argparse
+import os
+
+from snappl.logger import Lager
+from snappl.config import Config
+from snappl.image import OpenUniverse2024FITSImage
+
+pd.options.mode.chained_assignment = None  # default='warn'
```
wmwv, line 38: Explain why this line is needed and what it's doing.

OneObjFunc.py
```diff
@@ -0,0 +1,327 @@
...
+                   lc_start, lc_end, do_xshift, bg_gal_flux, do_rotation, airy,
+                   mismatch_seds, deltafcn_profile, noise, check_perfection,
+                   avoid_non_linearity):
+
+    Lager.debug(f'ID: {ID}')
+    psf_matrix = []
+    sn_matrix = []
+    cutout_wcs_list = []
+    im_wcs_list = []
+    # This is a catch for when I'm doing my own simulated WCS's
```
wmwv, line 57: `im_wcs_list = []` is the catch or `util_ref = None` is the catch.

Also, `WCSs` or `set of WCS`.  No possessive apostrophe here.

OneObjFunc.py
```diff
@@ -0,0 +1,327 @@
...
+        # TODO: Calculate peak MJD outside of the function
+        images, cutout_wcs_list, im_wcs_list, err, snra, sndec, ra, dec, \
+            exposures, object_type = fetchImages(num_total_imgs,
+                                                 num_detect_imgs, ID,
+                                                 sn_path, band, size,
+                                                 fit_background,
+                                                 roman_path,
+                                                 lc_start=lc_start,
+                                                 lc_end=lc_end)
+        if len(exposures[~exposures['DETECTED']]) == 0 and object_type == 'SN':
```
wmwv, line 74: I would break this into two lines

```
num_predection_images = len(exposures[~exposures['DETECTED']]) == 0
if num_predection_images == 0 and object_type == 'SN'
```

OneObjFunc.py
```diff
@@ -0,0 +1,327 @@
...
+            Lager.warning('No pre-detection images found in time range ' +
+                            'provided, skipping this object.')
+            return None
+
+        if num_total_imgs != np.inf and len(exposures) != num_total_imgs:
```
wmwv, line 79: Why would `num_total_imgs` be infinite?
Why are both of these checks necessary?

OneObjFunc.py
```diff
@@ -0,0 +1,327 @@
...
+                                             stampsize=size, x_center=snx,
+                                             y_center=sny, sed=sed,
+                                             photOps=source_phot_ops)
+            else:
+                stamp = galsim.Image(size, size, wcs=cutout_wcs_list[i])
+                profile = galsim.DeltaFunction()*sed
+                profile = profile.withFlux(1, roman_bandpasses[band])
+                convolved = galsim.Convolve(profile, sim_psf)
+                array = convolved.drawImage(roman_bandpasses[band],
+                                            method='no_pixel',
```
wmwv, line 237: Consider making this hardcoded a keyword to the function with a default so that it's accessible, even if it's never used.

OneObjFunc.py
```diff
@@ -0,0 +1,327 @@
+# TODO -- remove these next few lines!
```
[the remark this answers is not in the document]
ColeFMeldorf: File was just getting a little long but I certainly can!

simulation.py
```diff
@@ -64,10 +65,12 @@ def simulate_images(testnum, detim, ra, dec, do_xshift, do_rotation,
     galdec = dec + 1.5e-5
 
     if sim_lc is None:
-        if detim == 0:
+        # Here, if the user has not provided a light curve that they want
```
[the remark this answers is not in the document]
ColeFMeldorf: Offsetting the simulated galaxy from the supernova. Realizing that definitely shouldn't be hardcoded, I'll add an argument!

simulation.py
```diff
@@ -174,10 +180,9 @@ def simulate_images(testnum, detim, ra, dec, do_xshift, do_rotation,
     images = imagelist
     Lager.debug(f'images shape: {images[0].shape}')
     Lager.debug(f'images length {len(images)}')
-
-    util_ref = roman_utils(config_file='./temp_tds.yaml',
-                                       visit=base_pointing,
-                                       sca=base_sca)
+    file_path = str(pathlib.Path(__file__).parent/'temp_tds.yaml')
```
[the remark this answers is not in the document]
ColeFMeldorf: I just stole some of Rob's code that did this and it used str, but I guess you're right it's redundant!

simulation.py
```diff
@@ -64,10 +65,12 @@ def simulate_images(testnum, detim, ra, dec, do_xshift, do_rotation,
...
 
     if sim_lc is None:
-        if detim == 0:
+        # Here, if the user has not provided a light curve that they want
+        # simulated, we generate a default one.
+        if num_detect_imgs == 0:
             sim_lc = 0
         else:
-            d = np.linspace(5, 20, detim)
+            d = np.linspace(5, 20, num_detect_imgs)
```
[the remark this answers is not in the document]
ColeFMeldorf: Admittedly these are arbitrary values that I use to generate an exponential curve for a default simulated supernova light curve. I picked 5 and 20 because after I take an exponent they ended up with values in the range of fluxes of supernovae in the simulations. I could make them arguments instead, but that seems somewhat redundant when the user can already submit their own lightcurve?

OneObjFunc.py
ColeFMeldorf: Managed to actually fix this warning so this line removed.

OneObjFunc.py
ColeFMeldorf: My logic was that num_total_images was a cap on the total number of images that the algorithm would pull (if you were worried about computing time, say) so np.inf meant no cap on images. I could make it None instead? I just had set np.inf as the default when nothing was passed.

simulation.py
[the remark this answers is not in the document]
ColeFMeldorf: Alternatively I could just write out the fluxes that this produces, though I fear that would seem just as arbitrary!

AllASPFuncs.py
```diff
@@ -1041,21 +1041,21 @@ def slice_plot(fileroot):
         snx, sny = wcs.toImage(snra, sndec, units = 'deg')
 
         plt.subplot(len(cutout_wcs_list)//3 + 1,3,i+1)
-        if i >= testnum - detim:
-            plt.title('MagBias: ' + str(np.round(magresiduals[i - testnum + detim],4)) + ' mag')
+        if i >= num_total_imgs - num_detect_imgs:
```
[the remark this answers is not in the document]
ColeFMeldorf: Yes, but the code sorts all of the images by detected vs. non detected, so they will always be at the end!

ColeFMeldorf:
All changes addressed with 2 quick questions, let me know what you think!

OneObjFunc.py
wmwv: Separate from the motivation, I was wondering what the purpose of checking if `num_total_images != np.inf` and then check against `len(exposures)`.  The second will never return inf, so the first check seems like it's covered by the first.

wmwv: Yes, I think `None` instead of `np.inf` would be a bit more Pythonic, but only more clear if you rename the variable to refer to something like `max_num_total_images`.  But you don't have to rework that here.

ASP_tests.py
```diff
@@ -22,8 +22,8 @@
 import warnings
 import yaml
 
-warnings.simplefilter('ignore', category=AstropyWarning)
-warnings.filterwarnings("ignore", category=ErfaWarning)
+#warnings.simplefilter('ignore', category=AstropyWarning)
```
wmwv, line 25: Why are these here and commented-out?

RomanASP.py
```diff
@@ -80,10 +79,10 @@ def main():
                         help='Roman filter')
     parser.add_argument('-s', '--SNID', type=int, required=True,
                         help='Supernova ID')
-    parser.add_argument('-t', '--testnum', type=int, required=False,
+    parser.add_argument('-t', '--num_total_imgs', type=int, required=False,
```
wmwv, line 82: `num_total_images`

RomanASP.py
```diff
@@ -80,10 +79,10 @@ def main():
...
                         help='Number of images to use', default=np.inf)
     # TODO:change all instances of this variable to tot_images
-    parser.add_argument('-d', '--detim', type=int, required=False,
+    parser.add_argument('-d', '--num_detect_imgs', type=int, required=False,
```
wmwv, line 85: `num_detect_images`

RomanASP.py
```diff
@@ -167,8 +167,7 @@ def main():
         assert single_grid_point
     if avoid_non_linearity:
         assert deltafcn_profile
-    assert detim <= testnum
-    ####
+    assert num_detect_imgs <= num_total_imgs
```
wmwv, line 170: `num_detect_images` and `num_total_images`

wmwv (approved):
Minor comments

AllASPFuncs.py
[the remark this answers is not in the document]
wmwv: Do you have a test that verifies this?

PR #56, opened 2025-05-12
Added ability to do batch runs

ColeFMeldorf:
1. [x] Code now has the ability to run multiple SNe by listing multiple in the command line.
2. [x] Code can also run on multiple SNe by pointing to a csv file of SN IDs.
3. [x] Wrote a convenience function, get_SNe_from_parq, to get lists of SN IDs to run on.
4. [x] Wrote a test for this function.

### DIFF
--- a/AllASPFuncs.py
+++ b/AllASPFuncs.py
@@ -358,7 +358,7 @@ def findAllExposures(snid, ra, dec, peak, start, end, band, maxbg=24,
 
     res = res.loc[res['filter'] == band]
     det = res.loc[(res['date'] >= start) & (res['date'] <= end)]
-    det['offpeak_time'] = det['date'] - peak
+    det['offpeak_time'] = np.abs(det['date'] - peak) # this needs to be deleted
     det = det.sort_values('offpeak_time')
     if lc_start != -np.inf or lc_end != np.inf:
         det = det.loc[(det['offpeak_time'] >= lc_start) &
--- a/ASP_tests.py
+++ b/ASP_tests.py
@@ -63,15 +63,15 @@ def test_findAllExposures():
 
 
 def test_simulate_images():
-    band = 'F184'
     lam = 1293  # nm
+    band = 'F184'
     airy = \
         galsim.ChromaticOpticalPSF(lam, diam=2.36, aberrations=galsim.roman.
                                    getPSF(1, band, pupil_bin=1).aberrations)
-    images, im_wcs_list, cutout_wcs_list, psf_storage, sn_storage = \
+    images, im_wcs_list, cutout_wcs_list, sim_lc, util_ref = \
         simulate_images(testnum=10, detim=5, ra=7.541534306163982,
                         dec=-44.219205940734625, do_xshift=True,
-                        do_rotation=True, supernova=[10, 100, 1000,
+                        do_rotation=True, sim_lc=[10, 100, 1000,
                                                      10**4, 10**5],
                         noise=0, use_roman=False, band='F184',
                         deltafcn_profile=False, roman_path=roman_path, size=11,
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ On other systems 'conda' may be already in your path. Consult the documentation 
 ```
 module load conda
 ```
-### Create our conda environment.  
+### Create our conda environment.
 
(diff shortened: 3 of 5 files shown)


AllASPFuncs.py
```diff
@@ -443,13 +442,13 @@ def find_parq(ID, path, obj_type = 'SN'):
 
     for f in files:
         pqfile = int(f.split('_')[1].split('.')[0])
-        df = open_parq(pqfile, path, obj_type = obj_type)
-        #The issue is SN parquets store their IDs as ints and star parquets as strings.
+        df = open_parq(pqfile, path, obj_type=obj_type)
+        # The issue is SN parquets store their IDs as ints and star parquets as strings.
         # Should I convert the entire array or is there a smarter way to do this?
```
wmwv, line 447: Create an Issue to sort out the ID type confusion.

AllASPFuncs.py
```diff
@@ -1450,4 +1449,38 @@ def prep_data_for_fit(images, err, sn_matrix, wgt_matrix):
...
+
+    # Get the supernovae IDs from the parquet file
+    df = open_parq(parquet_file, sn_path, obj_type='SN')
+    # For now, this is only supported for SNe. TODO
+    if mag_limits is not None:
+        min_mag, max_mag = mag_limits
+        # This can't always be just g band I think. TODO
+        df = df[(df['peak_mag_g'] >= min_mag) & (df['peak_mag_g'] <= max_mag)]
+    SNID = df.id.values
+    SNID = SNID[np.log10(SNID) < 8]  # The 9 digit SNID SNe are weird for
```
wmwv, line 1474: This needs to stop.  These hardcoded 8 and 9 are going to blow up as soon as we try to run this on real data.  File an issue immediately to sort this out.

ColeFMeldorf: Yes, you're right, I acknowledge this wasn't smart, I just had no other way to tell them apart without very slowly checking every truth file. In the meantime, I am not sure if this is a good fix, but I've made a hotfix where the user just passes in an object type. In the spirit of being less reliant on the sims, this is the only way I can think of doing it when we don't know the object type ahead of time.  I do see that this disallows the edge case where a user wants to fit stars and SNe simultaneously, but I don't see that happening in all likelihood...? Fully open to other ideas though!

AllASPFuncs.py
```diff
@@ -1450,4 +1449,38 @@ def prep_data_for_fit(images, err, sn_matrix, wgt_matrix):
     wgt_matrix = np.array(wgt_matrix)
     wgt_matrix = np.hstack(wgt_matrix)
 
-    return images, err, sn_matrix, wgt_matrix
\ No newline at end of file
+    return images, err, sn_matrix, wgt_matrix
+
+def get_SNe_from_parq(parquet_file, sn_path, output_path, mag_limits=None):
```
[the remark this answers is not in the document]
ColeFMeldorf: Yes, I can name it that!

AllASPFuncs.py
```diff
@@ -1450,4 +1449,38 @@ def prep_data_for_fit(images, err, sn_matrix, wgt_matrix):
...
+    '''
+    Convenience function for getting a list of SNIDs that obey some conditions
+    from a parquet file. This is not used anywhere in the main algorithm.
+
+    Inputs:
+    parquet_file: the path to the parquet file
+    sn_path: the path to the supernova data
+    mag_limits: a tuple of (min_mag, max_mag) to filter the SNe by
```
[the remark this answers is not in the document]
ColeFMeldorf: The purpose of this is so that you could get a subset of the parquet file based on the brightness of the supernovae. For instance, Lauren/I show that the photometry errors blow up around 26th magnitude due to shot noise. I figured you might want to be able to exclude supernovae that were that dim when testing the algorithm.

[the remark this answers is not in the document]
ColeFMeldorf: And yes, peak magnitude does come from the truth files BUT this function is not used anywhere in the algorithm proper, only to prep a list of SNe to run on, so I think it's ok.....?

AllASPFuncs.py
ColeFMeldorf: Issue made!

PR #61, opened 2025-05-19
Snappl refactor

ColeFMeldorf:
In this PR, we:
1. [x] Refactor ConstructImages so that it returns Image objects
2. [x] However, the rest of the code does not use Image objects yet. So, temporarily, I turn the Image objects back into the numpy arrays that the rest of the code expects. 
Passes all tests!

### DIFF
--- a/AllASPFuncs.py
+++ b/AllASPFuncs.py
@@ -543,7 +543,7 @@ def gaussian(x, A, mu, sigma):
 
 
 def constructImages(exposures, ra, dec, size=7, background=False,
-                    roman_path=None):
+                    roman_path=None, truth='simple_model'):
 
     '''
     Constructs the array of Roman images in the format required for the linear algebra operations
@@ -562,8 +562,6 @@ def constructImages(exposures, ra, dec, size=7, background=False,
     '''
 
     bgflux = []
-    truth = 'simple_model'
-
     image_list = []
     cutout_image_list = []
 
--- a/AllASPFuncs.py
+++ b/AllASPFuncs.py
@@ -62,7 +62,7 @@ Adapted from code by Pedro Bernardinelli
 
 
                  _____  __     ___  __________
-                / __/ |/ /    / _ \/  _/_  __/
+                / __/ |/ /    / _ \/  _/_  __/image.
                _\ \/    /    / ___// /  / /
               /___/_/|_/    /_/  /___/ /_/
 
@@ -564,6 +564,9 @@ def constructImages(exposures, ra, dec, size=7, background=False,
     wcs_list = []
     truth = 'simple_model'
 
+    image_list = []
+    cutout_image_list = []
+
     Lager.debug(f'truth in construct images: {truth}')
 
     for indx, i in enumerate(exposures):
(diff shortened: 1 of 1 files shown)


AllASPFuncs.py
```diff
@@ -578,42 +576,12 @@ def constructImages(exposures, ra, dec, size=7, background=False,
...
-            err_cutout = 1 / img
-
-        else:
-            errordata, = image.get_data(which='noise')
-            err_cutout = Cutout2D(errordata, pixel, size, mode='strict').data
-
-
-        im = cutout
+            # In the future, I'd like to manually insert an array of ones for
+            # the error, or something.
```
wmwv, line 584: Why?

AllASPFuncs.py
```diff
@@ -626,6 +594,8 @@ def constructImages(exposures, ra, dec, size=7, background=False,
         im = cutout * zero
         '''
 
+        # Maybe repackage all of this stuff into a function? TODO
```
wmwv, line 597: This is background subtraction.  Whole suite of understanding of how to do this, so definitely .
"TODO: repackage all ..."

AllASPFuncs.py
```diff
@@ -643,20 +613,17 @@ def constructImages(exposures, ra, dec, size=7, background=False,
 
         bgflux.append(bg)
 
-        #If we are not fitting the background we manually subtract it here.
+        # If we are not fitting the background we manually subtract it here.
```
wmwv, line 616: "# If we are not fitting the background we use the value in SKY_MEAN

Also, I think you could more clearly restructure the use of `background` to be `subtract_background`.  But now that I mention that I wonder if it's `subtract_background` or `fit_background.

I'm actually kind of confused by these lines and the documentation for what `background` does.

AllASPFuncs.py
```diff
@@ -643,20 +613,17 @@ def constructImages(exposures, ra, dec, size=7, background=False,
...
         if not background and not truth == 'truth':
-            im -= image._get_header()['SKY_MEAN']
+            image_cutout._data -= image_cutout._get_header()['SKY_MEAN']
```
wmwv, line 618: I think it would be clearer to make the if above when you calculate `bg`, and then this data array operation can just simply be

```
image_cutout._data -= bg
Lager.debug(...)
```

AllASPFuncs.py
```diff
@@ -753,14 +720,31 @@ def fetchImages(num_total_images, num_detect_images, ID, sn_path, band, size, fi
...
+    cutout_wcs_list = []
+    im_wcs_list = []
+    err = []
+    for cutout, image in zip(cutout_image_list, image_list):
+        images.append(cutout._data)
+        cutout_wcs_list.append(galsim.AstropyWCS(wcs=cutout._wcs))
+        im_wcs_list.append(galsim.AstropyWCS(wcs=image._wcs))
+        err.append(cutout._noise)
+
     Lager.debug(f'here {np.shape(images)}')
```
wmwv, line 742: "here"?

wmwv (changes requested):
1. Document return values for `constructImages`
2. Start new Issue to capture the TODO flagged (and maybe one more) about `background`.

AllASPFuncs.py
ColeFMeldorf: Well my logic was that the truth images have no error, so their error arrays should be uniform?

AllASPFuncs.py
ColeFMeldorf: You're right. I've added some extra comments to make it clearer exactly why the if statements are here, but I will also create an issue to rename background ---> fit_background

AllASPFuncs.py
ColeFMeldorf: Issue created!

AllASPFuncs.py
wmwv: 1. Why 1?  Why not 0 or inf?  Would the error array be scaled to truth?  Is it an error array or a weight array?  
2. What's the intended use of a `truth` for `constructImages`?  How is it supposed to be passed?  Am I correct that right now it's just defined as `truth = "simple_model"` here as a line in this function?
3. Is the value of `truth` the same for all iterations?  I think so?  In that case, you can check for it before the loop.
4. Make an Issue to capture all of these thoughts, your ideas, and what should be done next.  Mark this comment with TODO.

AllASPFuncs.py
wmwv: Why is this marked as "resolved"?  My question was why does the `Lager.debug` statement say "here"  (fine, fine, that's not in this PR).  Saying "here" isn't super useful, unless you have only one message -- and if that's the case you still don't need to say "here" because they're only one place..  What you really want is for the logger framework to tell you where the error was emitted.

wmwv (commented):
Perhaps these comments are premature.  I'll wait a bit more for you to mark this as ready for re-review.

AllASPFuncs.py
ColeFMeldorf: TODO. I raised the issue just now!

ColeFMeldorf:
Ready for re-review!

wmwv (approved):
LGTM

PR #65, opened 2025-05-20
Renamed background and added a calculate bg function

ColeFMeldorf:
1. [x] Renamed `background` to `fit_background` for clarity.
2. [x] Pulled all of the background calculation steps into their own function, `calculate_background_level`
3. [x] Added a test for this function.

Passes all tests!

### DIFF
--- a/ASP_tests.py
+++ b/ASP_tests.py
@@ -256,3 +256,19 @@ def test_plot_lc():
     assert output[3][0] == 23.164154309471726
     assert output[4] == 182.088
     assert output[5] == 0.0
+
+
+def test_calculate_background_level():
+    from AllASPFuncs import calculate_background_level
+    test_data = np.ones((12, 12))
+    test_data[5:7, 5:7] = 1000
+    test_data[0:2, 0:12:2] = 10
+    test_data[-3:-1, 0:12:2] = 10
+    test_data[0:12:2, 0:2] = 10
+    test_data[0:12:2, -1:-3] = 10 # Adding some noise
+    # Expected output
+    expected_output = 1
+
+    output = calculate_background_level(test_data)
+    msg = f"Expected {expected_output}, but got {output}"
+    assert output == expected_output, msg
--- a/ASP_tests.py
+++ b/ASP_tests.py
@@ -185,6 +185,7 @@ def test_regression():
     config['turn_grid_off'] = False
     config['size'] = 19
     config['weighting'] = True
+    config['subtract_background'] = True
     # Weighting is a Gaussian width 1000 when this was made
     # In the future, this should be True, but random seeds not working rn.
     config['source_phot_ops'] = False
--- a/AllASPFuncs.py
+++ b/AllASPFuncs.py
@@ -200,7 +200,6 @@ def generateGuess(imlist, wcslist, ra_grid, dec_grid):
     return all_vals/len(wcslist)
 
 
-
 def construct_psf_background(ra, dec, wcs, x_loc, y_loc, stampsize, bpass,
                              use_roman, color=0.61, psf=None, pixel=False,
                              include_photonOps=False, util_ref=None,
@@ -542,7 +541,7 @@ def gaussian(x, A, mu, sigma):
     return A*np.exp(-(x-mu)**2/(2*sigma**2))
 
 
-def constructImages(exposures, ra, dec, size=7, background=False,
+def constructImages(exposures, ra, dec, size=7, fit_background=False,
                     roman_path=None, truth='simple_model'):
 
     '''
@@ -551,7 +550,8 @@ def constructImages(exposures, ra, dec, size=7, background=False,
     Inputs:
     exposures is a list of exposures from findAllExposures
     ra,dec: the RA and DEC of the SN
-    background: whether to subtract the background from the images
+    fit_background: If True, the background level is fit as a free parameter
+            in the forward modelling. Otherwise, we subtract it here.
     roman_path: the path to the Roman data
 
     Returns:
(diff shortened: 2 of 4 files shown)


ASP_tests.py
```diff
@@ -255,4 +255,20 @@ def test_plot_lc():
...
+
+
+def test_calculate_background_level():
+    from AllASPFuncs import calculate_background_level
+    test_data = np.ones((12, 12))
+    test_data[5:7, 5:7] = 1000
+    test_data[0:2, 0:12:2] = 10
+    test_data[-3:-1, 0:12:2] = 10
+    test_data[0:12:2, 0:2] = 10
+    test_data[0:12:2, -1:-3] = 10 # Adding some noise
```
wmwv, line 268: Why is this "noise"?

AllASPFuncs.py
```diff
@@ -639,6 +625,34 @@ def constructImages(exposures, ra, dec, size=7, background=False,
...
+
+    Returns:
+    bg, float, the estimated background level.
+
+    '''
+    size = im.shape[0]
+    bgarr = np.concatenate((im[0:size//4, 0:size//4].flatten(),
+                            im[0:size, size//4:size].flatten(),
+                            im[size//4:size, 0:size//4].flatten(),
+                            im[size//4:size, size//4:size].flatten()))
```
wmwv, line 644: What does this step do?  I don't understand why it's necessary.

AllASPFuncs.py
```diff
@@ -639,6 +625,34 @@ def constructImages(exposures, ra, dec, size=7, background=False,
     return cutout_image_list, image_list
 
 
+def calculate_background_level(im):
+    '''
+    A function for naively estimating the background level from a given image.
+    This may be replaced by a more sophisticated function later.
```
wmwv, line 631: Document what the current naive method does to estimate the background level.

AllASPFuncs.py
```diff
@@ -639,6 +625,34 @@ def constructImages(exposures, ra, dec, size=7, background=False,
...
+    Returns:
+    bg, float, the estimated background level.
+
+    '''
+    size = im.shape[0]
+    bgarr = np.concatenate((im[0:size//4, 0:size//4].flatten(),
+                            im[0:size, size//4:size].flatten(),
+                            im[size//4:size, 0:size//4].flatten(),
+                            im[size//4:size, size//4:size].flatten()))
+    bgarr = bgarr[bgarr != 0]
```
wmwv, line 645: Why is the value of `0` special?

AllASPFuncs.py
```diff
@@ -551,7 +550,8 @@ def constructImages(exposures, ra, dec, size=7, background=False,
     Inputs:
     exposures is a list of exposures from findAllExposures
     ra,dec: the RA and DEC of the SN
-    background: whether to subtract the background from the images
+    fit_background: If True, the background level is fit as a free parameter
```
wmwv, line 553: Think of a name of the variable in the context of the function.

E.g., for me `fit_background` means the opposite of how it's used in this function.

I think you're currently naming this variable based on what happens in a different function.

AllASPFuncs.py
```diff
@@ -598,34 +598,20 @@ def constructImages(exposures, ra, dec, size=7, background=False,
...
-            bgarr = bgarr[bgarr != 0]
-            if len(bgarr) == 0:
-                med = 0
-                bg = 0
-            else:
-                pc = np.percentile(bgarr, 84)
-                med = np.median(bgarr)
-                bgarr = bgarr[bgarr < pc]
-                bg = np.median(bgarr)
+            bg = calculate_background_level(imagedata)
```
wmwv, line 612: "manually"?

AllASPFuncs.py
```diff
@@ -598,34 +598,20 @@ def constructImages(exposures, ra, dec, size=7, background=False,
...
 
         # If we are not fitting the background we manually subtract it here.
-        if background:
-            # When background is true, we are including the background level as
+        if fit_background:
+            # When fit_background is true, we are including the background level as
             # a free parameter in our fit, so it should not be subtracted here.
             bg = 0
-        elif not background and not truth == 'truth':
+        elif not fit_background and not truth == 'truth':
```
wmwv, line 605: 1. I would nest this if, so that the `if truth` cases are just within the `elif not fit_background` block.
2. I would also just skip the `if fit_background:` part and replace that with set `bg = 0`, and then `if not fit_background:`.

AllASPFuncs.py
ColeFMeldorf: To clarify, this is leftover Pedro code. However, it is my understanding that this is taking the corners of the image where there is less likely to be contamination from the supernova in order to determine the background. However, upon closer inspection, it seems like there are typos that means it's selecting most of the image. I believe where it says `size//4:size`, it should read `3*size//4:size`. Would that make sense>

AllASPFuncs.py
ColeFMeldorf: Admittedly, I am not sure, it's a Pedro holdover. I'll remove it and see if anything breaks.

AllASPFuncs.py
wmwv: Don't rely on variable names from a person who's complaining they don't understand what the variable is used for :)

AllASPFuncs.py
ColeFMeldorf: Does not seem to matter.

AllASPFuncs.py
wmwv: I think that if you find yourself writing `if not foo:` as the main test (as essentially suggested by another comment I made), then use inverting that to be `if bar:` where `bar = not foo` likely makes sense.

ASP_tests.py
ColeFMeldorf: Because I am sigma clipping (now explicitly stated in the documentation) if the background is all the same number, all of the array gets cut when I try to cut values greater than 2 sigma above mean, because the mean is 1, 68th percentile is 1, 84th percentile is 1, etc. When that happened it would return nan thanks to no data. Now I am adding some other numbers so that the 84th percentile is > 1, the tens get sigma clipped and the ones are left behind!

AllASPFuncs.py
wmwv: Ah!  It just clicked, I understand what background is supposed to be doing.  Yes, `subtract_background` is the variable name I would suggest.

I think `constructImages` has an opportunity for a future name improvement, but we can leave that for later.

ColeFMeldorf:
All comments addressed or clarification requested!

AllASPFuncs.py
wmwv: "To clarify, this is leftover Pedro code."
It's all your code now :)

AllASPFuncs.py
wmwv: Write out the explanation for what this function does, and then questions such as what regions to calculate the background from and why 0 is special or not should be come more clear.

ASP_tests.py
wmwv: Perhaps "outliers" rather than "noise".  To me "noise" naively means some random sample from a normal (gaussian) distribution.  And add the explanation you just wrote to me above to the code.

AllASPFuncs.py
ColeFMeldorf: Point taken, a poor workman blames his tools. 
I am 90% sure what I said above is what he meant so I made that change!

AllASPFuncs.py
ColeFMeldorf: Sounds good! And point taken re: variable names haha

ColeFMeldorf:
all comments addressed!

wmwv (approved):
LGTM

PR #66, opened 2025-05-20
Refactor makeGrid

ColeFMeldorf:
My next step in snappl refactoring is the function that creates the background galaxy grid!
Unfortunately, I started and immediately realized the code I've written for it is atrocious (at least I'm learning...?)
So, this refactor might be a bit more involved. Therefore, I've written some tests for the makeGrid functions before I dive into taking them apart. In this PR:
1. [x] I write 3 tests for the 3 different grid methods I've come up with (regular, adaptive, and contour).
2. [x] Removed some variable names that aren't used.
3. [x] Added some documentation to makeGrid. This is not entirely done, I want to see exactly how I end up refactoring as then I'll know all the inputs and outputs.

By making this PR, I am drawing some attention to some pretty bad code. I know it's bad, so for this PR I want to focus on the tests so I can go in and fix the bad code. Thanks! :)

### DIFF
--- a/ASP_tests.py
+++ b/ASP_tests.py
@@ -268,9 +268,9 @@ def test_regular_grid():
     for key in wcs.keys():
         wcs[key] = wcs[key].item()
     wcs = galsim.wcs.readFromFitsHeader(wcs)[0]
-    ra_grid, dec_grid = local_grid(ra_center, dec_center, wcs, npoints=8,
+    ra_grid, dec_grid = local_grid(ra_center, dec_center, wcs,
                                    size=25, spacing=3.0, image=None,
-                                   spline_grid=False, percentiles=[],
+                                   percentiles=[],
                                    makecontourGrid=False)
     test_ra = np.array([7.67363133, 7.67373506, 7.67383878, 7.67355803,
                         7.67366176, 7.67376548, 7.67348473, 7.67358845,
@@ -293,9 +293,9 @@ def test_adaptive_grid():
     wcs = galsim.wcs.readFromFitsHeader(wcs)[0]
     compare_images = np.load('tests/testdata/images.npy')
     image = compare_images[:11**2].reshape(11, 11)
-    ra_grid, dec_grid = local_grid(ra_center, dec_center, wcs, npoints=8,
+    ra_grid, dec_grid = local_grid(ra_center, dec_center, wcs,
                                    size=11, spacing=3.0, image=image,
-                                   spline_grid=False, percentiles=[99],
+                                   percentiles=[99],
                                    makecontourGrid=False)
     test_ra = [7.67356034, 7.67366407, 7.67376779, 7.67348704, 7.67358874,
                7.67357896, 7.67360257, 7.67359279, 7.67369449, 7.67341373,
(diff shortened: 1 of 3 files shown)


ASP_tests.py
```diff
@@ -259,6 +259,76 @@ def test_plot_lc():
...
+    ra_center = wcs['CRVAL1']
+    dec_center = wcs['CRVAL2']
+    for key in wcs.keys():
+        wcs[key] = wcs[key].item()
+    wcs = galsim.wcs.readFromFitsHeader(wcs)[0]
+    ra_grid, dec_grid = local_grid(ra_center, dec_center, wcs,
+                                   size=25, spacing=3.0, image=None,
+                                   percentiles=[],
+                                   makecontourGrid=False)
+    test_ra = np.array([7.67363133, 7.67373506, 7.67383878, 7.67355803,
```
wmwv, line 275: This might be a bit more readable if you picked a "central" RA value and then added in offsets around it with respect to ra_central and dec_central.
```
ra_central = 7.673
test_ra = ra_center + np.array([0.00063133, 0.00073506, ...])
```
But this is a pretty straightforward test case.  No need to change.  Just an idea to consider.

ASP_tests.py
```diff
@@ -259,6 +259,76 @@ def test_plot_lc():
...
+                        7.67366176, 7.67376548, 7.67348473, 7.67358845,
+                        7.67369218])
+    test_dec = np.array([-44.26396874, -44.26391831, -44.26386787,
+                         -44.26389673, -44.26384629, -44.26379586,
+                         -44.26382471, -44.26377428, -44.26372384])
+    assert np.allclose(ra_grid, test_ra, rtol=1e-7), "RA vals do not match"
```
wmwv, line 281: I think here you want an absolute tolerance.  Because RA=0 or Dec=0 are perfectly valid values.  It doesn't matter for this test, but I wanted to suggest it to you so you start thinking about atol vs rtol.

ASP_tests.py
```diff
@@ -259,6 +259,76 @@ def test_plot_lc():
...
 
 
+def test_regular_grid():
+    from AllASPFuncs import local_grid
+    wcs = np.load('./tests/testdata/wcs_dict.npz', allow_pickle=True)
+    wcs = dict(wcs)
+    ra_center = wcs['CRVAL1']
+    dec_center = wcs['CRVAL2']
+    for key in wcs.keys():
+        wcs[key] = wcs[key].item()
```
wmwv, line 269: What does this do?  Why is it necessary?

ASP_tests.py
```diff
@@ -272,4 +342,5 @@ def test_calculate_background_level():
 
     output = calculate_background_level(test_data)
     msg = f"Expected {expected_output}, but got {output}"
-    assert output == expected_output, msg
\ No newline at end of file
+    assert output == expected_output, msg
```
wmwv, line 345: I think you're comparing a float here.  I suggest using a numpy test close rather than exact equality.

AllASPFuncs.py
```diff
@@ -82,14 +83,15 @@ def local_grid(ra_center, dec_center, wcs, npoints, size = 25, spacing = 1.0, im
     difference = int((size - subsize)/2)
 
     x_center, y_center = wcs.toImage(ra_center, dec_center, units='deg')
+    x = np.arange(difference, subsize+difference, spacing)
```
wmwv, line 86: You could write this as
`x = difference + np.arange(0, subsize, spacing)`
which is arguably a little more compact and helps communicate more clearly what's being set up.

wmwv (changes requested):
Minor comments.  More suggestions to think about in the future rather than requests for this PR.  The float comparison one maybe is important to fix.

ASP_tests.py
ColeFMeldorf: Good point, that makes sense.

ASP_tests.py
ColeFMeldorf: For some reason the way I saved the WCS it saves the actual data and the dtype in a tuple, which galsim WCS does not know how to read. This is me just pulling out the actual data.

ASP_tests.py
ColeFMeldorf: Fixed!

ASP_tests.py
wmwv: Very good.  Add those two sentences you just wrote here to the code next time you're near it.

PR #70, opened 2025-05-22
Refactor make grid, Part 2

ColeFMeldorf:
1. [x] The overly complicated, multipurpose function `local_grid` has been split into 3 functions, `regular_grid`, `make_adaptive_grid`, and `make_contour_grid`.
2. [x] The tests written for local_grid have been switched so that they now run on these new functions, which still pass.
3. [x] Added documentation (with some fun schematics!) for these functions.
4. [x] Need to update the README
5. [x] Need to change `regular_grid` ---> `make_regular_grid` for symmetry.

### DIFF
--- a/ASP_tests.py
+++ b/ASP_tests.py
@@ -268,10 +268,8 @@ def test_regular_grid():
     for key in wcs.keys():
         wcs[key] = wcs[key].item()
     wcs = galsim.wcs.readFromFitsHeader(wcs)[0]
-    ra_grid, dec_grid = local_grid(ra_center, dec_center, wcs,
-                                   size=25, spacing=3.0, image=None,
-                                   percentiles=[],
-                                   makecontourGrid=False)
+    ra_grid, dec_grid = regular_grid(ra_center, dec_center, wcs,
+                                   size=25, spacing=3.0)
     test_ra = np.array([7.67363133, 7.67373506, 7.67383878, 7.67355803,
                         7.67366176, 7.67376548, 7.67348473, 7.67358845,
                         7.67369218])
@@ -293,34 +291,24 @@ def test_adaptive_grid():
     wcs = galsim.wcs.readFromFitsHeader(wcs)[0]
     compare_images = np.load('tests/testdata/images.npy')
     image = compare_images[:11**2].reshape(11, 11)
-    ra_grid, dec_grid = local_grid(ra_center, dec_center, wcs,
-                                   size=11, spacing=3.0, image=image,
-                                   percentiles=[99],
-                                   makecontourGrid=False)
-    test_ra = [7.67356034, 7.67366407, 7.67376779, 7.67348704, 7.67358874,
-               7.67357896, 7.67360257, 7.67359279, 7.67369449, 7.67341373,
-               7.67351746, 7.67362119]
-    test_dec = [-44.26425446, -44.26420403, -44.26415359, -44.26418244,
-                -44.26414017, -44.26413057, -44.26413345, -44.26412385,
-                -44.26408158, -44.26411043, -44.26405999, -44.26400956]
-    assert np.allclose(ra_grid, test_ra, rtol=1e-7), "RA vals do not match"
-    assert np.allclose(dec_grid, test_dec, rtol=1e-7), "Dec vals do not match"
+    ra_grid, dec_grid = make_adaptive_grid(ra_center, dec_center, wcs,
+                                           image=image, percentiles=[99])
+    test_ra = [7.67356034, 7.67359491, 7.67362949, 7.67366407, 7.67369864,]
+    test_dec = [-44.26425446, -44.26423765, -44.26422084, -44.26420403, -44.26418721]
+    assert np.allclose(ra_grid[:5], test_ra, rtol=1e-7), "RA vals do not match"
+    assert np.allclose(dec_grid[:5], test_dec, rtol=1e-7), "Dec vals do not match"
 
 
 def test_contour_grid():
     from AllASPFuncs import local_grid
     wcs = np.load('./tests/testdata/wcs_dict.npz', allow_pickle=True)
     wcs = dict(wcs)
-    ra_center = wcs['CRVAL1']
-    dec_center = wcs['CRVAL2']
     for key in wcs.keys():
         wcs[key] = wcs[key].item()
     wcs = galsim.wcs.readFromFitsHeader(wcs)[0]
     compare_images = np.load('tests/testdata/images.npy')
     image = compare_images[:11**2].reshape(11, 11)
-    ra_grid, dec_grid = local_grid(ra_center, dec_center, wcs,
-                                   size=11, spacing=3.0, image=image,
-                                   makecontourGrid=True)
+    ra_grid, dec_grid = contourGrid(image, wcs)
     test_ra = [7.67356034, 7.67359491, 7.67362949, 7.67366407]
     test_dec = [-44.26425446, -44.26423765, -44.26422084, -44.26420403]
     msg = "RA vals do not match"
--- a/ASP_tests.py
+++ b/ASP_tests.py
@@ -275,8 +275,8 @@ def test_make_regular_grid():
     test_dec = np.array([-44.26396874, -44.26391831, -44.26386787,
                          -44.26389673, -44.26384629, -44.26379586,
                          -44.26382471, -44.26377428, -44.26372384])
-    assert np.allclose(ra_grid, test_ra, rtol=1e-7), "RA vals do not match"
-    assert np.allclose(dec_grid, test_dec, rtol=1e-7), "Dec vals do not match"
+    assert np.allclose(ra_grid, test_ra, atol=1e-7), "RA vals do not match"
+    assert np.allclose(dec_grid, test_dec, atol=1e-7), "Dec vals do not match"
 
 
 def test_make_adaptive_grid():
@@ -293,8 +293,8 @@ def test_make_adaptive_grid():
                                            image=image, percentiles=[99])
     test_ra = [7.67356034, 7.67359491, 7.67362949, 7.67366407, 7.67369864,]
     test_dec = [-44.26425446, -44.26423765, -44.26422084, -44.26420403, -44.26418721]
-    assert np.allclose(ra_grid[:5], test_ra, rtol=1e-7), "RA vals do not match"
-    assert np.allclose(dec_grid[:5], test_dec, rtol=1e-7), "Dec vals do not match"
+    assert np.allclose(ra_grid[:5], test_ra, atol=1e-7), "RA vals do not match"
+    assert np.allclose(dec_grid[:5], test_dec, atol=1e-7), "Dec vals do not match"
 
 
 def test_make_contour_grid():
@@ -309,22 +309,22 @@ def test_make_contour_grid():
     test_ra = [7.67356034, 7.67359491, 7.67362949, 7.67366407]
     test_dec = [-44.26425446, -44.26423765, -44.26422084, -44.26420403]
     msg = "RA vals do not match"
-    assert np.allclose(ra_grid[:4], test_ra, rtol=1e-7), msg
+    assert np.allclose(ra_grid[:4], test_ra, atol=1e-7), msg
     msg = "Dec vals do not match"
-    assert np.allclose(dec_grid[:4], test_dec, rtol=1e-7), msg
+    assert np.allclose(dec_grid[:4], test_dec, atol=1e-7), msg
 
 
 def test_calculate_background_level():
     from AllASPFuncs import calculate_background_level
     test_data = np.ones((12, 12))
     test_data[5:7, 5:7] = 1000
-    test_data[0:2, 0:12:2] = 10
-    test_data[-3:-1, 0:12:2] = 10
-    test_data[0:12:2, 0:2] = 10
-    test_data[0:12:2, -1:-3] = 10  # Adding some noise
-    # Expected output
-    expected_output = 1
+    test_data[0:2, 0:12:2] = 123
+    test_data[-3:-1, 0:12:2] = 123
+    test_data[0:12:2, 0:2] = 123
+    test_data[0:12:2, -1:-3] = 123  # Adding some outliers to prevent all of
+    # the data from being sigma clipped.
 
+    expected_output = 1
     output = calculate_background_level(test_data)
     msg = f"Expected {expected_output}, but got {output}"
     assert np.isclose(output, expected_output, rtol=1e-7), msg
--- a/ASP_tests.py
+++ b/ASP_tests.py
@@ -179,7 +179,7 @@ def test_regression():
     config['use_roman'] = True
     config['use_real_images'] = True
     config['fetch_SED'] = False
-    config['makecontourGrid'] = True
+    config['contour_grid'] = True
     config['band'] = 'Y106'
     config['adaptive_grid'] = True
     config['turn_grid_off'] = False
@@ -259,7 +259,7 @@ def test_plot_lc():
     assert output[5] == 0.0
 
 
-def test_regular_grid():
+def test_make_regular_grid():
     wcs = np.load('./tests/testdata/wcs_dict.npz', allow_pickle=True)
     wcs = dict(wcs)
     ra_center = wcs['CRVAL1']
@@ -267,7 +267,7 @@ def test_regular_grid():
     for key in wcs.keys():
         wcs[key] = wcs[key].item()
     wcs = galsim.wcs.readFromFitsHeader(wcs)[0]
-    ra_grid, dec_grid = regular_grid(ra_center, dec_center, wcs,
+    ra_grid, dec_grid = make_regular_grid(ra_center, dec_center, wcs,
                                    size=25, spacing=3.0)
     test_ra = np.array([7.67363133, 7.67373506, 7.67383878, 7.67355803,
                         7.67366176, 7.67376548, 7.67348473, 7.67358845,
@@ -279,7 +279,7 @@ def test_regular_grid():
     assert np.allclose(dec_grid, test_dec, rtol=1e-7), "Dec vals do not match"
 
 
-def test_adaptive_grid():
+def test_make_adaptive_grid():
     wcs = np.load('./tests/testdata/wcs_dict.npz', allow_pickle=True)
     wcs = dict(wcs)
     ra_center = wcs['CRVAL1']
@@ -297,7 +297,7 @@ def test_adaptive_grid():
     assert np.allclose(dec_grid[:5], test_dec, rtol=1e-7), "Dec vals do not match"
 
 
-def test_contour_grid():
+def test_make_contour_grid():
     wcs = np.load('./tests/testdata/wcs_dict.npz', allow_pickle=True)
     wcs = dict(wcs)
     for key in wcs.keys():
@@ -305,7 +305,7 @@ def test_contour_grid():
     wcs = galsim.wcs.readFromFitsHeader(wcs)[0]
     compare_images = np.load('tests/testdata/images.npy')
     image = compare_images[:11**2].reshape(11, 11)
-    ra_grid, dec_grid = contourGrid(image, wcs)
+    ra_grid, dec_grid = make_contour_grid(image, wcs)
     test_ra = [7.67356034, 7.67359491, 7.67362949, 7.67366407]
     test_dec = [-44.26425446, -44.26423765, -44.26422084, -44.26420403]
     msg = "RA vals do not match"
@@ -321,7 +321,7 @@ def test_calculate_background_level():
     test_data[0:2, 0:12:2] = 10
     test_data[-3:-1, 0:12:2] = 10
     test_data[0:12:2, 0:2] = 10
-    test_data[0:12:2, -1:-3] = 10 # Adding some noise
+    test_data[0:12:2, -1:-3] = 10  # Adding some noise
     # Expected output
     expected_output = 1
 
--- a/ASP_tests.py
+++ b/ASP_tests.py
@@ -159,7 +159,7 @@ def test_run_on_star():
     config_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
                                'config.yaml')
     config = yaml.safe_load(open(config_path))
-    config['turn_grid_off'] = True
+    config['grid_type'] = 'none'
 
     with tempfile.NamedTemporaryFile(mode="w", suffix=".yaml", delete=False)\
             as temp_config:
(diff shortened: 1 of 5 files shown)


AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
+                pass
+            elif num == 1:
+                xes.append(y)
+                ys.append(x)  #Why this? TODO
+            else:
+                xx = np.linspace(x - 0.6, x + 0.6, num+2)[1:-1]
+                yy = np.linspace(y - 0.6, y + 0.6, num+2)[1:-1]
+                X, Y = np.meshgrid(xx, yy)
+                ys.extend(list(X.flatten()))
+                xes.extend(list(Y.flatten())) #Why this? TODO
```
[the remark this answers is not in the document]
ColeFMeldorf: Wow I'm glad I accidentally clicked on this actually

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
+                      subsize = 9):
     '''
-    Generates a local grid around a RA-Dec center, choosing step size and
-    number of points
+    Generates a regular grid around a RA-Dec center, choosing step size.
+
+    ra_center, dec_center: floats, coordinate center of the image
+    wcs: the WCS of the image, currently a galsim.fitswcs.AstropyWCS object
+    spacing: int, spacing of grid points in pixels.
+    subsize: If you don't care as much about the edges of the image, you can
```
wmwv, line 81: "If you don't care as much about the edges of the image".  Are these the first 12 words you want someone to read about the parameter?  Document in order of concept and importance.  Justify example reasons toward the end.  Start with one the parameter controls.

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
+    xx, yy = np.meshgrid(x+1, y+1)
+    xx = xx.flatten()
+    yy = yy.flatten()
+    Lager.debug(f'Built a grid with {np.size(xx)} points')
 
-        xx = np.array(xes)
-        yy = np.array(ys)
+    result = wcs.toWorld(xx, yy, units='deg')
+    ra_grid = result[0]
+    dec_grid = result[1]
```
wmwv, line 110:
```
ra_grid, dec_grid = wcs.toWorld(xx, yy, units='deg')
```

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
 
-    else:
-        xx, yy = np.meshgrid(x+1, y+1)
+def make_adaptive_grid(ra_center, dec_center, wcs,
+                       image, percentiles=[45, 90], subsize=9):
     '''
-        subsize = 8 #Taking a smaller square inside the image to fit on
-        difference = int((size - subsize)/2)
+    Construct an "adaptive grid" which allocates model grid points to model
+    the background galaxy according to the brightness of the image.
```
wmwv, line 118: Brightness with respect to what?

Sky background level?
Image variance?
Absolute nJy?

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
+    # We need to make sure that the image is not zero, otherwise we get
+    # infinities in the log space.
+    imcopy[imcopy <= 0] = 1e-10
+    imcopy = np.log(imcopy)
+    bins = [0]
+    bins.extend(np.nanpercentile(imcopy, percentiles))
+    bins.append(100)
+    Lager.debug(f'BINS: {bins}')
+
+    a = np.digitize(imcopy, bins)
```
wmwv, line 167: Pick a more informative name than `a`.

P.S.  Nice use of `np.digitize`

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
+    ys = []
+
+    yvals = np.rint(y).astype(int)
+    xvals = np.rint(x).astype(int)
+    for xindex in xvals:
+        x = xindex + 1
+        for yindex in yvals:
+            y = yindex + 1
+            # xindex and yindex are the indices within the numpy array, while
+            # x and y are the actual locations in pixel space.
```
wmwv, line 178: You're carrying around this default assumption of how to refer to x, y coordinates in pixel space.  There are different conventions.  Three reasonable conventions of the center of the "first" pixel being defined at (0, 0), (0.5, 0.5), (1, 1)).  I think you think that one of them is "right".  But that's just one choice.  Be explicit about your assumption/choice.

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
+            # This used to be x and y in here:
+            num = int(a[xindex][yindex])
+            if num == 0:
+                pass
+            elif num == 1:
+                xes.append(y)
+                ys.append(x)  #Why this? TODO
+            else:
+                xx = np.linspace(x - 0.6, x + 0.6, num+2)[1:-1]
```
wmwv, line 189: What is 0.6?

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
+    bins = [0]
+    bins.extend(np.nanpercentile(imcopy, percentiles))
+    bins.append(100)
+    Lager.debug(f'BINS: {bins}')
+
+    a = np.digitize(imcopy, bins)
+    xes = []
+    ys = []
+
+    yvals = np.rint(y).astype(int)
```
wmwv, line 171: Explain what this block is making.

AllASPFuncs.py
```diff
@@ -859,42 +888,45 @@ def getWeights(cutout_wcs_list, size, snra, sndec, error=None,
...
-
-    returns:
+    Returns:
     ra_grid, dec_grid: numpy arrays of floats of the ra and dec locations for
                     model grid points.
-
-
     '''
     if adaptive_grid:
+        if contour_grid:
```
wmwv, line 912: This is a confusing nesting.  I think you can simplify to make this a one level if/elif/else construct.

ASP_tests.py
```diff
@@ -336,7 +321,7 @@ def test_calculate_background_level():
     test_data[0:2, 0:12:2] = 10
     test_data[-3:-1, 0:12:2] = 10
     test_data[0:12:2, 0:2] = 10
-    test_data[0:12:2, -1:-3] = 10 # Adding some noise
+    test_data[0:12:2, -1:-3] = 10  # Adding some noise
```
wmwv, line 324: Same comment as in previous PR.  Call this "outlier" and make the value not `10` to make it clearly distinct from the levels you're setting in the lines above.

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
+    the grid on.
+    percentiles: list of floats, the percentiles to use to bin the image. The
+                more bins, the more possible grid points could be placed in
+                that pixel. For instance, say if you had bins [45, 90],
+                as is default. A pixel that lies in the 30th percentile for
+                brightness would get 1 point. A pixel at the 50th percentiile
+                for brightness would get a 2x2 grid of points.
+                A pixel above the 90th percentile would get a 3x3 grid of
+                points. If you have more bins, you could go even higher to
+                4x4 and 5x5 etc. These points are evenly paced within the
```
wmwv, line 133: "placed"

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
+    ra_center, dec_center: floats, coordinate center of the image
+    wcs: the WCS of the image, currently a galsim.fitswcs.AstropyWCS object
+    image: 2D numpy array of floats of shape (size x size), the image to build
+    the grid on.
+    percentiles: list of floats, the percentiles to use to bin the image. The
+                more bins, the more possible grid points could be placed in
+                that pixel. For instance, say if you had bins [45, 90],
+                as is default. A pixel that lies in the 30th percentile for
+                brightness would get 1 point. A pixel at the 50th percentiile
+                for brightness would get a 2x2 grid of points.
```
wmwv, line 130: Is the number points a pixel gets determined by the order of the percentiles?  

[45, 90] -> [1x1, 2x2]
[10, 25, 50, 75, 90] -> [1x1, 2x2, 3x3, 4x4, 5x5]

What happens if the list of floats is not in ascending order?
[45, 90, 10] -> ???

How many images do you need to provide constraints on a 5x5 subpixel grid?

wmwv (changes requested):
Explain the sub-pixelization, particularly with binning more clearly.

Explain what those for loops are doing.

Do you want the variables to be boolean adaptive_grid vs contour_grid
or a single configuration parameter that can be a string that specifies which one to use or None?

4. Sit down and explain to yourself pixel conventions for image arrays.  General choices, and choices in FITS, numpy, sextractor.  Explain how WCS definition works in the context of x, y zero convention.  Then explain this to someone else.  Then come back to this code and more clearly address convention choices and places where you've inserted a `+1`.

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
+    the background galaxy according to the brightness of the image.
 
-        spacing = 1.0
-        x = np.arange(difference, subsize+difference, spacing)
-        y = np.arange(difference, subsize+difference, spacing)
+    Inputs:
+    ra_center, dec_center: floats, coordinate center of the image
+    wcs: the WCS of the image, currently a galsim.fitswcs.AstropyWCS object
+    image: 2D numpy array of floats of shape (size x size), the image to build
+    the grid on.
```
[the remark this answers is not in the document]
ColeFMeldorf: Done!

AllASPFuncs.py
ColeFMeldorf: I think I meant spaced but done!

AllASPFuncs.py
ColeFMeldorf: Done!

AllASPFuncs.py
ColeFMeldorf: Thanks! and changed it to `brightness_levels`

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
+    imcopy[imcopy <= 0] = 1e-10
+    imcopy = np.log(imcopy)
+    bins = [0]
+    bins.extend(np.nanpercentile(imcopy, percentiles))
+    bins.append(100)
+    Lager.debug(f'BINS: {bins}')
+
+    a = np.digitize(imcopy, bins)
+    xes = []
+    ys = []
```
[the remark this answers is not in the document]
ColeFMeldorf: Done!

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
-        result = wcs.toWorld(xx, yy, units='deg')
-        ra_grid = result[0]
-        dec_grid = result[1]
-    else:
-        Lager.warning('swapped x and y here')
-        result = wcs.pixel_to_world(yy, xx)
-        ra_grid = result.ra.deg
-        dec_grid = result.dec.deg
-
+    result = wcs.toWorld(xx, yy, units='deg')
```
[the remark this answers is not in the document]
ColeFMeldorf: done!

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
 
-def local_grid(ra_center, dec_center, wcs, size=25, spacing=1.0,
-               image=None, percentiles=[], makecontourGrid=True):
-
+def make_regular_grid(ra_center, dec_center, wcs, size, spacing=1.0,
+                      subsize = 9):
     '''
-    Generates a local grid around a RA-Dec center, choosing step size and
-    number of points
+    Generates a regular grid around a RA-Dec center, choosing step size.
```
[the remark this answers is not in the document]
ColeFMeldorf: check

AllASPFuncs.py
```diff
@@ -70,107 +70,134 @@
...
+
+    Returns:
+    ra_grid, dec_grid: 1D numpy arrays of floats, the RA and DEC of the grid.
+
+
     '''
-    Lager.debug('image shape: {}'.format(np.shape(image)))
-    # Build the basic grid
-    subsize = 9  # Taking a smaller square inside the image to fit on
+    Lager.debug('GRID TYPE: REGULARLY SPACED')
```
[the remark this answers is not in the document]
ColeFMeldorf: No Idea, got too excited I guess. I've made them normal case.

AllASPFuncs.py
ColeFMeldorf: Done!

AllASPFuncs.py
ColeFMeldorf: Added an explanation.

README.md
```diff
@@ -85,8 +85,8 @@ initial WCS. |
...
 |---------------------|-------|-------------------------------------------------------------------------------------------------------------------------------------------|
 | pixel               | bool  | If true, use a pixel (tophat) function rather than a delta function to be convolved with the PSF in order to build the model.            |
-| makecontourGrid     | bool  | A new method being developed to generate the adaptive grid. Seems to be better! TODO: Consider replacing the default method.             |
-| fit_background      | bool  | If true, add an extra parameter that fits for the mean sky background level. Should be false since the exact number is in the image header. |
+| contour_grid     | bool  | A new method being developed to generate the adaptive grid. Seems to be better! TODO: Consider replacing the default method.             |
+| subtract_background      | bool  | If False, add an extra parameter that fits for the mean sky background level. If true, subtract the SKYMEAN from the image header. |
```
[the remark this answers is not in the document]
ColeFMeldorf: Yes, I'll remove it and re add it in another PR.

ASP_tests.py
ColeFMeldorf: done!

ASP_tests.py
```diff
@@ -293,34 +289,23 @@ def test_adaptive_grid():
...
-    test_dec = [-44.26425446, -44.26420403, -44.26415359, -44.26418244,
-                -44.26414017, -44.26413057, -44.26413345, -44.26412385,
-                -44.26408158, -44.26411043, -44.26405999, -44.26400956]
-    assert np.allclose(ra_grid, test_ra, rtol=1e-7), "RA vals do not match"
-    assert np.allclose(dec_grid, test_dec, rtol=1e-7), "Dec vals do not match"
+    ra_grid, dec_grid = make_adaptive_grid(ra_center, dec_center, wcs,
+                                           image=image, percentiles=[99])
+    test_ra = [7.67356034, 7.67359491, 7.67362949, 7.67366407, 7.67369864,]
+    test_dec = [-44.26425446, -44.26423765, -44.26422084, -44.26420403, -44.26418721]
+    assert np.allclose(ra_grid[:5], test_ra, rtol=1e-7), "RA vals do not match"
```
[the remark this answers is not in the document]
ColeFMeldorf: done!

AllASPFuncs.py
ColeFMeldorf: Yes, I explained it, in the code:
>    subpixel_grid_width: When we place the model points in a pixel, we place
                        them on a small range of locations within the pixel.
                        For instance, 0.25, 0.5, and 0.75 for x values. However
                        I've found this leads to awkward gaps in grid points
                        between pixels. For instance, the point at 0.25 would
                        be half a pixel from the point located at 0.75 in the
                        next lower pixel, and only 0.25 from the point at 0.5
                        in the same pixel.
                        Therefore, subpixel_grid_width can be
                        set to something larger than 1 so that the evenly
                        spaced points are spaced out more, reducing these gaps.
                        For instance, if you set to 1.2, the x values of the
                        pixels would be [0.2, 0.5, 0.8] instead, reducing
                        inter-pixel gaps.

and getting ahead hopefully getting ahead of your next comment, I made it so that it is no longer hardcoded!

AllASPFuncs.py
ColeFMeldorf: This is in absolute counts, which I believe is in photoelectrons (or, equivalently, ADU, since I was told the gain is 1?). If that is correct, I can include that information here.

ColeFMeldorf:
Other than doing a deep dive on understanding the difference between indexing rules for FITs, numpy, etc., all comments have been addressed or responded with a question. I'll do that reading and think through all of the +1s but I just wanted to get the rest back to you first.

ASP_tests.py
```diff
@@ -293,53 +289,42 @@ def test_adaptive_grid():
...
-    test_data[0:2, 0:12:2] = 10
-    test_data[-3:-1, 0:12:2] = 10
-    test_data[0:12:2, 0:2] = 10
-    test_data[0:12:2, -1:-3] = 10 # Adding some noise
-    # Expected output
-    expected_output = 1
+    test_data[0:2, 0:12:2] = 123
+    test_data[-3:-1, 0:12:2] = 123
+    test_data[0:12:2, 0:2] = 123
+    test_data[0:12:2, -1:-3] = 123  # Adding some outliers to prevent all of
```
wmwv, line 324: Ahh, I didn't understand this before.  This comment applies to all 4 of these lines, not just the last one.

Can you move the comment to before these 4 lines to more clearly indicate the it applies to all 4.  I had previously thought that the comment only applied to the last line; that's why I asked for the value 10 to be different than the other 10s, because I thought there were different things.

AllASPFuncs.py
```diff
@@ -70,107 +70,172 @@
...
+                      subsize=9):
     '''
-    Generates a local grid around a RA-Dec center, choosing step size and
-    number of points
+    Generates a regular grid around a (RA, Dec) center, choosing step size.
+
+    ra_center, dec_center: floats, coordinate center of the image
+    wcs: the WCS of the image, currently a galsim.fitswcs.AstropyWCS object
+    spacing: int, spacing of grid points in pixels.
+    subsize: Int, width of the grid in pixels.
```
wmwv, line 81: lower case 'int' for consistency

AllASPFuncs.py
```diff
@@ -70,107 +70,172 @@
...
+
     '''
-    Lager.debug('image shape: {}'.format(np.shape(image)))
-    # Build the basic grid
-    subsize = 9  # Taking a smaller square inside the image to fit on
+    if subsize > size:
+        Lager.warning('subsize is larger than the image size. ' +
+                      f'{size} > {subsize}. This would cause model points to' +
+                      ' be placed outside the image. Reducing subsize to' +
+                      ' match the image size.')
```
wmwv, line 99: Hmmm.. if you're not going to allow subsize > image size, then conceptually aren't you requiring that all the points are within the circle that inscribes the square to handle all potential rotations?

I.e. saying that subsize == size is fine, makes the first image and its orientation special.  You know that you are creating grid points that will not be constrained by other images at other rotations.

AllASPFuncs.py
wmwv: My point is that absolute counts don't mean anything.  The exposure time could be 1s or 1000s for the same scene and you would have 1000x difference in flux.  

I think that what you're actually doing is calculating things based on the relative brightness of pixel values (you're calculating percentiles).  If that's the case then clarify this documentation here to talk about relative brightness.

AllASPFuncs.py
```diff
@@ -70,107 +70,172 @@
...
+    imcopy = np.log(imcopy)
+    bins = [0]
+    bins.extend(np.nanpercentile(imcopy, percentiles))
+    bins.append(100)
+    Lager.debug(f'BINS: {bins}')
+
+    brightness_levels = np.digitize(imcopy, bins)
+    xs = []
+    ys = []
+    # Round y and x locations to the nearest pixel. This is necessary because
```
wmwv, line 204: Nearest pixel edge, corner, center?

AllASPFuncs.py
```diff
@@ -70,107 +70,172 @@
...
+                        between pixels. For instance, the point at 0.25 would
+                        be half a pixel from the point located at 0.75 in the
+                        next lower pixel, and only 0.25 from the point at 0.5
+                        in the same pixel.
+                        Therefore, subpixel_grid_width can be
+                        set to something larger than 1 so that the evenly
+                        spaced points are spaced out more, reducing these gaps.
+                        For instance, if you set to 1.2, the x values of the
+                        pixels would be [0.2, 0.5, 0.8] instead, reducing
+                        inter-pixel gaps.
```
wmwv, line 163: I think I finally understand what `subpixel_grid_width` does and why it's defined this way.  But I still think this is confusing.  I think that subsampling should automatically fill in pixels with equal-area spacing of the points  (1/6, 3/6 5/6) in 1D.

AllASPFuncs.py
wmwv: Which way do you want to round?

AllASPFuncs.py
```diff
@@ -70,107 +70,172 @@
...
+            num = int(brightness_levels[xindex][yindex])
+            if num == 0:
+                pass
+            elif num == 1:
+                xs.append(y)
+                ys.append(x)  # I know I swap this because Astropy takes (y,x)
+                # order but I'd really like to iron out all the places I do
+                # this rather than doing it so off the cuff. TODO
+            else:
+                xx = np.linspace(x - subpixel_grid_width/2,
```
wmwv, line 225:
```
xx = x + np.linspace(-subpixel_grid_width/2, +subpixel_grid_width/2, num+2)[1:-1]
```
Wait, why is this `num+2` and `[1:-1]` instead of `num` and the whole array?

AllASPFuncs.py
```diff
@@ -857,55 +924,60 @@ def getWeights(cutout_wcs_list, size, snra, sndec, error=None,
...
-                galra = ra_grid[0]
-                galdec = dec_grid[0]
-            else:
-                galra = ra_grid[106]
-                galdec = dec_grid[106]
-
-        ra_grid = np.array(ra_grid)
-        dec_grid = np.array(dec_grid)
+            raise NotImplementedError
+            # I need to figure out how to turn the single grid point test
```
wmwv, line 974: TODO -> Issue

wmwv (changes requested):
Assorted comments for future Cole.  Go ahead and merge this branch.

AllASPFuncs.py
ColeFMeldorf: Yes! I fully agree! This is why I ended up writing a new function a while ago called `make_contour_grid` which largely skirts this issue! It does not do a "look at each pixel one by one" thing like this code does. However, I chose not to delete this old method just in case. `make_adaptive_grid`, where this code is, is not my default grid method.

AllASPFuncs.py
ColeFMeldorf: done!

ASP_tests.py
ColeFMeldorf: Oh I see. Fixed!

AllASPFuncs.py
ColeFMeldorf: Those don't quite give the same result, because np.linspace is inclusive of the bounds. Consider:
`num = 3
subpixel_grid_width = 1.2
print(np.linspace(0.5 - subpixel_grid_width/2, 0.5 + subpixel_grid_width/2, num+2)[1:-1])
print(np.linspace(0.5 - subpixel_grid_width/2, 0.5 + subpixel_grid_width/2, num))
`

You get out:
`[0.2 0.5 0.8]
[-0.1  0.5  1.1]`

wmwv: Oh, I understand.  You're intentionally creating the end points and dropping them.  Dividing the same space in 1/num vs 1/(num+2) is not the same thing.  Thank you.

PR #76, opened 2025-05-25
Outputfile improvements

ColeFMeldorf:
Added a catch for negative fluxes.
sn_ra --> obj_ra.
Added band and zeropoint in the output file.

Unclear where to find the SIM_zeropoint?

### DIFF
--- a/ASP_tests.py
+++ b/ASP_tests.py
@@ -11,6 +11,7 @@ sys.path.insert(0, str(pathlib.Path(__file__).parent/"extern/snappl"))
 from AllASPFuncs import *
 from astropy.io import ascii
 from astropy.utils.exceptions import AstropyWarning
+from astropy.wcs import WCS
 from erfa import ErfaWarning
 from simulation import simulate_galaxy, simulate_images, simulate_supernova, \
                        simulate_wcs
@@ -300,7 +301,8 @@ def test_make_contour_grid():
     wcs = dict(wcs)
     for key in wcs.keys():
         wcs[key] = wcs[key].item()
-    wcs = galsim.wcs.readFromFitsHeader(wcs)[0]
+    #wcs = galsim.wcs.readFromFitsHeader(wcs)[0]
+    wcs = WCS(wcs)
     compare_images = np.load('tests/testdata/images.npy')
     image = compare_images[:11**2].reshape(11, 11)
     ra_grid, dec_grid = make_contour_grid(image, wcs)
@@ -329,3 +331,20 @@ def test_calculate_background_level():
     msg = f"Expected {expected_output}, but got {output}"
     assert np.isclose(output, expected_output, rtol=1e-7), msg
 
+
+def test_calc_mags_and_err():
+    flux = np.array([-1e2, 1e2, 1e3, 1e4])
+    sigma_flux = np.array([10, 10, 10, 10])
+    band = 'Y106'
+    mags, magerr, zp = calc_mags_and_err(flux, sigma_flux, band)
+
+    test_mags = np.array([666., 27.66165575,  25.16165575,  22.66165575])
+    test_magerr = np.array([6.6600000e+02, 1.0857362e-01,
+                            1.0857362e-02, 1.0857362e-03])
+    test_zp = 15.023547191066587
+
+    assert np.allclose(mags, test_mags, rtol=1e-7), \
+         "The magnitudes do not match"
+    assert np.allclose(magerr, test_magerr, rtol=1e-7), \
+         "The magnitude errors do not match"
+    assert np.allclose(zp, test_zp, rtol=1e-7), "The zeropoint does not match"
--- a/ASP_tests.py
+++ b/ASP_tests.py
@@ -205,7 +205,10 @@ def test_regression():
         # According to Michael and Rob, this is roughly what can be expected
         # due to floating point precision.
         msg = "The lightcurves do not match for column %s" % col
-        assert np.allclose(current[col], comparison[col], rtol=1e-7), msg
+        if col == 'band':
+            assert np.array_equal(current[col], comparison[col]), msg
+        else:
+            assert np.allclose(current[col], comparison[col], rtol=1e-7), msg
 
 
 def test_get_galsim_SED():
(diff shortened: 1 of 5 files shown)


AllASPFuncs.py
```diff
@@ -1449,11 +1449,21 @@ def calc_mags_and_err(flux, sigma_flux, band, zp = None):
...
-    return mags, magerr
+    positive = (flux > 0)
+    mags = np.zeros_like(flux)
+    magerr = np.zeros_like(flux)
+    mags[positive] = -2.5*np.log10(flux[positive]) + \
+    2.5*np.log10(exptime[band]*area_eff) + zp
+    magerr[positive] = 2.5 * sigma_flux[positive] / \
+        (flux[positive] * np.log(10))
+    mags[~positive] = 666
+    magerr[~positive] = 666 # 666 Matches SNANA convention
```
wmwv, line 1465: I strongly disagree.  It should be NaN.

AllASPFuncs.py
```diff
@@ -1449,11 +1449,21 @@ def calc_mags_and_err(flux, sigma_flux, band, zp = None):
...
     zp = roman.getBandpasses()[band].zeropoint if zp is None else zp
-    mags = -2.5*np.log10(flux) + 2.5*np.log10(exptime[band]*area_eff) + zp
-    magerr = 2.5 * sigma_flux / (flux * np.log(10))
-    return mags, magerr
+    positive = (flux > 0)
+    mags = np.zeros_like(flux)
+    magerr = np.zeros_like(flux)
+    mags[positive] = -2.5*np.log10(flux[positive]) + \
+    2.5*np.log10(exptime[band]*area_eff) + zp
+    magerr[positive] = 2.5 * sigma_flux[positive] / \
```
wmwv, line 1462: I prefer to write this as `(2.5 / np.log(10) * (sigma_flux / flux)` because it captures the two parts (i): the coefficient and the jacobian, and the (ii): signal-to-noise ratio.

ASP_tests.py
```diff
@@ -329,3 +329,20 @@ def test_calculate_background_level():
...
+    sigma_flux = np.array([10, 10, 10, 10])
+    band = 'Y106'
+    mags, magerr, zp = calc_mags_and_err(flux, sigma_flux, band)
+
+    test_mags = np.array([666., 27.66165575,  25.16165575,  22.66165575])
+    test_magerr = np.array([6.6600000e+02, 1.0857362e-01,
+                            1.0857362e-02, 1.0857362e-03])
+    test_zp = 15.023547191066587
+
+    assert np.allclose(mags, test_mags, rtol=1e-7), \
```
wmwv, line 344: atol makes more sense for mag and magerr

wmwv (changes requested):
I strongly object to the `666` magice number bad value convention.  log of a negative number is well-defined in IEEE 574 as NaN.  We should keep it that way.

AllASPFuncs.py
ColeFMeldorf: Sounds good, I'm not married to it, was a Rick suggestion. Fixed!

AllASPFuncs.py
ColeFMeldorf: Makes sense to me! Fixed

ASP_tests.py
ColeFMeldorf: Done!

ColeFMeldorf:
Note that in this latest commit I needed to change the snappl imports now that snappl is a package not a submodule.

ASP_tests.py
```diff
@@ -205,7 +205,10 @@ def test_regression():
         # According to Michael and Rob, this is roughly what can be expected
         # due to floating point precision.
         msg = "The lightcurves do not match for column %s" % col
-        assert np.allclose(current[col], comparison[col], rtol=1e-7), msg
+        if col == 'band':
```
wmwv, line 208: Add a quick note that `band` stores a string so we do an `array_equal` exact comparison instead of the `allclose` float comparison.

AllASPFuncs.py
```diff
@@ -1449,11 +1450,21 @@ def calc_mags_and_err(flux, sigma_flux, band, zp = None):
             'Y106': 302.275,
             'Z087': 101.7}
 
+    flux = np.atleast_1d(flux)
+    sigma_flux = np.atleast_1d(sigma_flux)
```
wmwv, line 1454: Why does `calc_mags_and_err` change scalar input to an array?  I would find this behavior surprising, particularly because it's not documented.

AllASPFuncs.py
```diff
@@ -1449,11 +1450,21 @@ def calc_mags_and_err(flux, sigma_flux, band, zp = None):
...
+
     area_eff = roman.collecting_area
     zp = roman.getBandpasses()[band].zeropoint if zp is None else zp
-    mags = -2.5*np.log10(flux) + 2.5*np.log10(exptime[band]*area_eff) + zp
-    magerr = 2.5 * sigma_flux / (flux * np.log(10))
-    return mags, magerr
+    positive = (flux > 0)
+    mags = np.zeros_like(flux)
```
wmwv, line 1459: Use singular `mag`.  As a general rule, don't use plurals with variable names because it's hard to remember which is singular and which is plural, and then you have `magerr` but not `magserr`.

AllASPFuncs.py
```diff
@@ -1449,11 +1450,21 @@ def calc_mags_and_err(flux, sigma_flux, band, zp = None):
...
+    magerr = np.zeros_like(flux)
+    mags[positive] = -2.5*np.log10(flux[positive]) + \
+    2.5*np.log10(exptime[band]*area_eff) + zp
+    magerr[positive] = (2.5 / np.log(10) * \
+        (sigma_flux[positive] / flux[positive]))
+    mags[~positive] = np.nan
+    magerr[~positive] = np.nan
```
wmwv, line 1466: You can make this much simpler.  Just calculate

```
mag = -2.5 * np.log10(flux) + 2.5*np.log10(exptime[band]*area_eff) + zp
magerr = (2.5 / np.log(10) * (sigma_flux / flux))
```

wmwv (changes requested):
1. Simplify mag calculation.
2. Rename `mags` to `mag`
3. Explain why `calc_mag_and_err` changes the input type from scalar to array or, my preference, don't do this.

AllASPFuncs.py
wmwv: I suppose this means I would want the function name to be `calc_mag_and_err`.

ASP_tests.py
ColeFMeldorf: done!

ColeFMeldorf:
All comments addressed, let me know what you think!

AllASPFuncs.py
```diff
@@ -1451,9 +1467,13 @@ def calc_mags_and_err(flux, sigma_flux, band, zp = None):
 
     area_eff = roman.collecting_area
     zp = roman.getBandpasses()[band].zeropoint if zp is None else zp
-    mags = -2.5*np.log10(flux) + 2.5*np.log10(exptime[band]*area_eff) + zp
-    magerr = 2.5 * sigma_flux / (flux * np.log(10))
-    return mags, magerr
+    positive = (flux > 0)
+    mag = np.zeros_like(flux)
+    magerr = np.zeros_like(flux)
```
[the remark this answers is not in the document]
ColeFMeldorf: check!

RomanASP.py
```diff
@@ -25,8 +25,8 @@
 from roman_imsim.utils import roman_utils
 import scipy.sparse as sp
 from simulation import simulate_images
-from snappl.logger import Lager
-from snappl.config import Config
+from snpit_utils.logger import SNLogger as Lager
```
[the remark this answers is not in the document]
ColeFMeldorf: I certainly can, though I do love the name! :)

PR #94, opened 2025-06-02
percentiles and spacing now arguments

ColeFMeldorf:
In this PR I:
1. [x] Add spacing and percentiles to the passable arguments in the config.yaml file
2. [x] Update the README with a description

### DIFF
--- a/AllASPFuncs.py
+++ b/AllASPFuncs.py
@@ -1718,6 +1718,7 @@ def run_one_object(ID, object_type, num_total_images, num_detect_images, roman_p
                    lc_start, lc_end, do_xshift, bg_gal_flux, do_rotation, airy,
                    mismatch_seds, deltafcn_profile, noise, check_perfection,
                    avoid_non_linearity, sim_gal_ra_offset, sim_gal_dec_offset,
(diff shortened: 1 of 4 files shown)


wmwv:
What's the relationship between the `perc_argument`, `percentiles_and_spacing`, and `percentiles` branches?

ColeFMeldorf:
The first time I made these changes I accidentally pushed them to main. I used git revert to get main back to where it was before the push. I then tried to make a PR on percentiles_and_spacing in order to commit my changes, but because that commit was in the history already, git thought main was ahead of percentiles_and_spacing, so I could not do a PR. I then git reflog-ed back to before the push to main to try to make a new branch from there (percentiles) but the same issue occurred. I tried a few more things but eventually reasoned this was too much work for 6 lines of code so I just made a new branch and rewrote my changes. That's this PR.

I know the way to fix this would probably have been to edit the history of main to entirely remove that commit, rather than just revert it. However, I remembered you told me I shouldn't use commit --amend after a push, and I figured this was the same, I shouldn't modify history after pushing. Let me know if that's wrong.

PR #112, opened 2025-07-02
Refactoring construct_psf_background to use snappl PSF objects

ColeFMeldorf:
In this PR, we now have `campari `using `snappl `PSFs for` construct_psf_background`. This required:
1. [x] Creating the `snappl `ou24psf class, see `snappl` PR 37 https://github.com/Roman-Supernova-PIT/snappl/pull/37
2. [x] Replacing my calls to galsim drawImage with `snappl `PSF get stamp methods
3. [x] `construct_psf_background `now uses the WCS of the SCA not the cutout, as that is the coordinates that `snappl `expects
4. [x] Because of this the `construct_psf_background `test had to be updated to get a whole `OpenUniverse2024FitsImage `instead of a saved cutout and wcs
5. [x] `test_lc.ecsv `needed to be updated because the old method of getting the PSFs to construct the background was slightly wrong. I am decently confident that this new way is more correct, though I do acknowledge the flux change is large.


campari/AllASPFuncs.py
```diff
@@ -305,21 +303,11 @@ def construct_psf_background(ra, dec, wcs, x_loc, y_loc, stampsize,
...
-    # With plus ones here I recover the values pre-refactor!
-
-    if psf is None:
-        # How different are these two methods? TODO XXX
-        pupil_bin = 8
-        # psf = util_ref.getPSF(x_loc, y_loc, pupil_bin=pupil_bin)
-        psf = galsim.roman.getPSF(1, band, pupil_bin=pupil_bin, wcs=galsim_wcs)
-
+    # I call this x_SCA to highlight that it's the location in the SCA, not the cutout.
+    x_SCA, y_SCA = sca_wcs.world_to_pixel(ra, dec)
```
wmwv, line 307: lower case.  `x_sca, y_sca`

campari/AllASPFuncs.py
```diff
@@ -305,21 +303,11 @@ def construct_psf_background(ra, dec, wcs, x_loc, y_loc, stampsize,
...
     bpass = roman.getBandpasses()[band]
 
-    psfs = np.zeros((stampsize * stampsize, np.size(x)))
+    psfs = np.zeros((stampsize * stampsize, np.size(x_SCA)))
```
wmwv, line 310: x_sca

campari/AllASPFuncs.py
```diff
@@ -332,19 +320,22 @@ def construct_psf_background(ra, dec, wcs, x_loc, y_loc, stampsize,
         point *= sed
 
     point = point.withFlux(1, bpass)
-    oversampling_factor = 1
-    convolvedpsf = galsim.Convolve(point, psf)
-    stamp = galsim.Image(stampsize*oversampling_factor,
-                         stampsize*oversampling_factor, wcs=galsim_wcs)
+
+    pointing = util_ref.visit
+    SCA = util_ref.sca
```
wmwv, line 325: lower case.  `sca = util_ref.sca`

campari/AllASPFuncs.py
```diff
@@ -332,19 +320,22 @@ def construct_psf_background(ra, dec, wcs, x_loc, y_loc, stampsize,
...
+
+    psf_object = PSF.get_psf_object("ou24PSF", pointing=pointing, sca=SCA, size=stampsize, include_photonOps=False)
+    # See run_one_object documentation to explain this pixel coordinate conversion.
+    x_loc = int(np.floor(x_loc + 0.5))
+    y_loc = int(np.floor(y_loc + 0.5))
+
     # Loop over the grid points, draw a PSF at each one, and append to a list.
-    for a, ij in enumerate(zip(x.flatten(), y.flatten())):
+    for a, ij in enumerate(zip(x_SCA.flatten(), y_SCA.flatten())):
```
wmwv, line 333: x_sca

campari/AllASPFuncs.py
```diff
@@ -271,16 +271,14 @@ def generateGuess(imlist, ra_grid, dec_grid):
...
                              util_ref=None, band=None):
 
     """Constructs the background model around a certain image (x,y) location
     and a given array of RA and DECs.
     Inputs:
     ra, dec: arrays of floats, RA and DEC values for the grid
-    wcs: the wcs of the image, if the image is a cutout, this MUST be the wcs
-        of the cutout. A snappl.wcs.BaseWCS object.
+    sca_wcs: the wcs of the entire image, i.e. the entire SCA. A snappl.wcs.BaseWCS object.
     x_loc, y_loc: floats,the pixel location of the image in the FULL image,
```
wmwv, line 282: More specifically, is this the pixel location of the center of the stamp in the FULL image?

campari/AllASPFuncs.py
```diff
@@ -332,19 +320,22 @@ def construct_psf_background(ra, dec, wcs, x_loc, y_loc, stampsize,
...
+    # See run_one_object documentation to explain this pixel coordinate conversion.
+    x_loc = int(np.floor(x_loc + 0.5))
+    y_loc = int(np.floor(y_loc + 0.5))
+
     # Loop over the grid points, draw a PSF at each one, and append to a list.
-    for a, ij in enumerate(zip(x.flatten(), y.flatten())):
+    for a, ij in enumerate(zip(x_SCA.flatten(), y_SCA.flatten())):
         if a % 50 == 0:
-            Lager.debug(f"Drawing PSF {a} of {np.size(x)}")
+            Lager.debug(f"Drawing PSF {a} of {np.size(x_loc)}")
```
wmwv, line 335: Above you were using size of `x_sca`, but here `x_loc`.  Why?  They should of course be the same, but it makes the reader thinks that there's perhaps a reason to use versus the other.  Consider define a new variable.  Something like:
`grid_length = len(x_sca)`  (or `np.size`, as you're using here instead of `len`, either is fine)

wmwv: Or `npoints`, which is the name used in the docstring.

campari/AllASPFuncs.py
wmwv, line 333: Just say `for a, (i, j) in enumerate(...`
and remove the
`i, j = ij` line below

wmwv: Although, I would expect `i` and `j` to be ints.  But I think they are floats.  So I would probably say

`for a, (x, y) in enumerate(...`
and use x, y down below in the `get_stamp` call.

campari/AllASPFuncs.py
```diff
@@ -1718,7 +1717,7 @@ def run_one_object(ID, ra, dec, object_type, exposures, num_total_images, num_de
         if grid_type != "none":
             background_model_array = \
                 construct_psf_background(ra_grid, dec_grid,
-                                         cutout_image_list[i].get_wcs(),
+                                         image_list[i].get_wcs(),
```
wmwv, line 1720: You defined `image_list[i].get_wcs` as a variable above, `whole_sca_wcs`, so use it again here.

campari/AllASPFuncs.py
```diff
@@ -271,16 +271,14 @@ def generateGuess(imlist, ra_grid, dec_grid):
     return all_vals/len(wcslist)
 
 
-def construct_psf_background(ra, dec, wcs, x_loc, y_loc, stampsize,
-                             psf=None, pixel=False,
+def construct_psf_background(ra, dec, sca_wcs, x_loc, y_loc, stampsize, psf=None, pixel=False,
```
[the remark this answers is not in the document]
ColeFMeldorf: Here background means the model of the static scene, but you're right we also use background to mean the sky background, i.e. the single value we subtract off of the entire image. 
I could define it explicitly or I could rename this 'construct_psf_scene_model' if that's clearer?

campari/AllASPFuncs.py
ColeFMeldorf: check

campari/AllASPFuncs.py
ColeFMeldorf: check

campari/AllASPFuncs.py
ColeFMeldorf: I did not know you could put a tuple in the for statement, thank you!

campari/AllASPFuncs.py
ColeFMeldorf: check!

campari/AllASPFuncs.py
ColeFMeldorf: No that's a typo, I meant location of the object. Thanks for the catch!

ColeFMeldorf:
All comments addressed except for the first one, which I just had a quick question on. Ready for re-review!

campari/AllASPFuncs.py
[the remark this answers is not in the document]
wmwv: After some more thought and better understanding, I think this function is misnamed in both `psf` and `background`.  This function doesn't return a PSF.  It returns an image, generated using a scene model and the PSF for each image (that's why the galsim call is to `drawImage`); although normalized to 1 (maybe?  I don't know what you get with galsim photon-shooting; it seems like it has to be normalized to ~num_photons).

I don't like `construct_*_scene_model`, because you're constructing the _image_ from the model.  The model is already constructed.  I appreciate that you're passing models that are flux normalized to 1 and then fitting for the coefficient (as part of the later matrix inversion), so it's not exactly the scene, it's like the basis vector the scene.  So none of (`_scene`, `_model`, `_image`) is quite right.  For me, I think _scene` is the best:

Same arguments for `construct_psf_source`.

So I would suggest that these functions should be called

`construct_psf_background` -> `construct_static_scene`  | `construct_scene_static`
`construct_psf_source` -> `construct_transient_scene` | `construct_scene_transient`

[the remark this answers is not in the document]
wmwv: Please push the changes.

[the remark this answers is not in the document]
wmwv: Ah, I see they've been pushed now.

PR #120, opened 2025-07-15
Image List now functional.

ColeFMeldorf:
In this PR we allow for an image list to be passed to findAllExposures. The code will then only use those images. To do so:

1. [x] The -i command line argument takes in a file and the code opens it.
2. [x] The number of maximum images to fetch is overridden to have no limit (otherwise the function would stop looking before it found all of the images)
3. [x]  Added a warning if not all of the images were found. NOTE: should this throw an error instead?

Note that the single quotes in line 323 are forced :))

Second Note: Technically, only a pointing is required, not a pointing and an SCA, because the SN cannot be in two different SCA in the same pointing. However, I assumed that whatever the program the user used to get the images they want to pass spits out pointing and SCA, hence why I assume the csv file has both pointings and SCAs in it. Is there a smarter way to do this?

### DIFF
--- a/campari/AllASPFuncs.py
+++ b/campari/AllASPFuncs.py
@@ -563,11 +563,11 @@ def constructImages(exposures, ra, dec, size=7, subtract_background=True,
 
     Lager.debug(f"truth in construct images: {truth}")
 
-    for indx, i in enumerate(exposures):
+    for indx, exp in enumerate(exposures):
         Lager.debug(f"Constructing image {indx} of {len(exposures)}")
-        band = i["BAND"]
-        pointing = i["pointing"]
-        sca = i["SCA"]
+        band = exp["BAND"]
+        pointing = exp["pointing"]
+        sca = exp["SCA"]
 
         # TODO : replace None with the right thing once Exposure is implemented
 
--- a/campari/RomanASP.py
+++ b/campari/RomanASP.py
@@ -168,9 +168,14 @@ def main():
     # used.  TODO: specify type of image, and adapt the code to handle
     # that.  Right now it will just assume openuniverse 2024.
 
-    parser.add_argument("-i", "--img_list", default=None,
-                        help="File with list of images. Note that if you pass an image list, the max number of images "
-                             "to use will be ignored, and campari will use all the images in the list.")
+    parser.add_argument(
+        "-i",
+        "--img_list",
+        default=None,
+        help="File with list of images. Note that if you pass an image list, the arguments "
+        "--max_no_transient_images and --max_transient_images"
+        " will be ignored, and campari will use all the images in the list.",
+    )
 
     ####################
     # What does it mean to run on stars??????  Assume constant flux? No


campari/RomanASP.py
```diff
@@ -267,7 +270,13 @@ def main():
                          "to run campari.")
 
     if args.img_list is not None:
-        raise NotImplementedError("--img-list not yet supported.")
+        columns = ["Pointing", "SCA"]
```
wmwv, line 273: I'd prefer lower-case for the column names.  Just so that all columns are lower-case and you never have to thinking about whether or not a name should be Upper, lower, ALLCAPS, or camelCase.

campari/RomanASP.py
```diff
@@ -267,7 +270,13 @@ def main():
                          "to run campari.")
 
     if args.img_list is not None:
```
wmwv, line 272: The name is `img-list` (dash rather than underscore) in the parser.  Test should have caught this.

campari/RomanASP.py
```diff
@@ -267,7 +270,13 @@ def main():
...
-        raise NotImplementedError("--img-list not yet supported.")
+        columns = ["Pointing", "SCA"]
+        image_df = pd.read_csv(args.img_list, header=None, names=columns)
+        # If provided a list, we want to make sure we continue searching until all the images are found. So we set:
+        max_no_transient_images = None
+        max_transient_images = None
```
wmwv, line 277: Makes sense.  Document this behavior in the parse --help mesage.

campari/RomanASP.py
```diff
@@ -231,6 +231,9 @@ def main():
     assert grid_type in ["regular", "adaptive", "contour",
                          "single", "none"], er
 
+    max_no_transient_images = num_total_images - num_detect_images  # This might be refactored in a more intelligent way
+    max_transient_images = num_detect_images                        # later.
```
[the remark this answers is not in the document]
wmwv: Consider renaming the inputs to the function to match these variable names.

campari/RomanASP.py
```diff
@@ -303,10 +312,16 @@ def main():
...
+                                         maxbg=max_no_transient_images,
+                                         maxdet=max_transient_images, return_list=True,
                                          band=band, image_selection_start=image_selection_start,
-                                         image_selection_end=image_selection_end)
+                                         image_selection_end=image_selection_end, pointing_list=image_df["Pointing"])
+
+            if not np.array_equiv(np.sort(exposures["Pointing"]), np.sort(image_df["Pointing"])):
+                Lager.warning("Unable to find the object in all the pointings in the image list. Specifically, the"
+                              " following pointings were not found: "
+                              f"{np.setdiff1d(image_df['Pointing'], exposures['Pointing'])}")
```
wmwv, line 323: "pointing"

campari/RomanASP.py
```diff
@@ -303,10 +312,16 @@ def main():
...
                                          roman_path=roman_path,
-                                         maxbg=num_total_images - num_detect_images,
-                                         maxdet=num_detect_images, return_list=True,
+                                         maxbg=max_no_transient_images,
+                                         maxdet=max_transient_images, return_list=True,
                                          band=band, image_selection_start=image_selection_start,
-                                         image_selection_end=image_selection_end)
+                                         image_selection_end=image_selection_end, pointing_list=image_df["Pointing"])
+
+            if not np.array_equiv(np.sort(exposures["Pointing"]), np.sort(image_df["Pointing"])):
```
wmwv, line 320: "pointing"

campari/RomanASP.py
wmwv: Why are these upper-case?  Why is there the upper-case in AllASPFuncs:L383?

And then you have to convert in line L402.

This case comes up in a bunch of places, so I see that it's not this PR that starts this, but really, I encourage you to keep everything lower-case to minimize mental burden (for others, even if it all makes sense to you).

And side, side note, I suggest using 'mjd' instead of 'date' for the column name.

campari/RomanASP.py
ColeFMeldorf: Fixed!

campari/RomanASP.py
wmwv: single quotes are totally fine here for exactly the reasons you articulate.

campari/RomanASP.py
ColeFMeldorf: check

campari/RomanASP.py
ColeFMeldorf: #123

ColeFMeldorf:
Ok this is ready for re-review!
Apologies for forgetting a test, I didn't add a new function so the single "if then" synapse in my brain for making a new test did not fire. Obviously it now makes sense that there should be a new test for new functionality.
Also I said this above but just to make it super clear: I lower cased all of the "Pointing" s,  but made an issue to go through and do the rest, column names, function names, etc.

campari/RomanASP.py
```diff
@@ -161,12 +164,13 @@ def main():
...
-    # If instead you give imglist, then you expliclty list the images
+    # If instead you give img_list, then you expliclty list the images
     # used.  TODO: specify type of image, and adapt the code to handle
     # that.  Right now it will just assume openuniverse 2024.
 
-    parser.add_argument("-i", "--img-list", default=None,
-                        help="File with list of images")
+    parser.add_argument("-i", "--img_list", default=None,
+                        help="File with list of images. Note that if you pass an image list, the max number of images "
+                             "to use will be ignored, and campari will use all the images in the list.")
```
wmwv, line 173: Specifically use the variable names exactly as spelled to make it clear to the user that you're talking about exactly the same thing.  Say "if ..., then max_no_transient_images and max_transient_images will be ignore, ..."

campari/AllASPFuncs.py
```diff
@@ -566,7 +566,7 @@ def constructImages(exposures, ra, dec, size=7, subtract_background=True,
     for indx, i in enumerate(exposures):
         Lager.debug(f"Constructing image {indx} of {len(exposures)}")
         band = i["BAND"]
-        pointing = i["Pointing"]
+        pointing = i["pointing"]
         sca = i["SCA"]
```
wmwv, line 570: Oh, I'd really prefer 'exp' or 'exposure" to the variable 'i' here.  In a for loop, 'i' would be assumed to be the index, not the value.

(Feel free to ignore, this is out of scope for the PR).

wmwv (changes requested):
1. Improve help message for --img_list.

campari/RomanASP.py
ColeFMeldorf: Check!

PR #137, opened 2025-08-04
Sim psf

ColeFMeldorf:
General goal for this PR is to get the simulation side of campari to use the new PSF and WCS objects so as to match the actual fitting pipeline. This one spiralled a bit.

1. [x] Because simulated SNe can be simulated on WCSs that do not appear in the OU24 sims, WCS is now a passable argument to the PSF object when getting the stamp. This is a snappl change, not a campari change, but it is integral to getting this PR to work. I checked, and passing the WCS does not cause regression tests to fail, so we're good there.
2. [x] A few np.atleast1d added because sometimes the grid is of size 1 for testing purposes
3. [x] Added ability to make a single grid point placed exactly at the galaxy, to test the code's ability to place bg points correctly.
4. [x] The base_pointing and base_sca that are used to simulate the WCS are now no longer hardcoded
5. [x] The half light radius of the bulge and disk of the simulated galaxy are now passable arguments
6. [x] The tests for the simulation code are now in a separate file, test_simulations.py


campari/AllASPFuncs.py
```diff
@@ -526,8 +532,7 @@ def construct_transient_scene(x, y, pointing, sca, stampsize=25, x_center=None,
 
     psf_object = PSF.get_psf_object("ou24PSF_slow", pointing=pointing, sca=sca,
```
rknop, line 533: I will need to go back and refamiliarize myself with the structure of camparias a whole here.

However, you will eventually not want to have a specific PSF instantiation function inside functions here.  Rather, you'll need to pass a PSF in.  This function shouldn't need to know what kind of PSF it has (which also means the next line, which is the actual change in this PR, will also have to change, as the input_wcs isn't a general part of PSF), but should just use the PSF.

campari/AllASPFuncs.py
```diff
@@ -889,6 +895,10 @@ def make_grid(grid_type, images, ra, dec, percentiles=[], spacing=0.75,
     make_exact: Currently not implemented, but will construct the grid in such
                 a way on a simulated image that the recovered model is accurate
                 to machine precision. TODO
+    sim_galra, sim_galdec: floats, the RA and DEC of a single simulated galaxy, only
```
rknop, line 898: I might rename these to `singlera` and `singledec`.

The use case you have right now is a simulated galaxy, but if you're using `grid_type="single"` for _any_ reason, then these two parameters become necessary.

campari/AllASPFuncs.py
```diff
@@ -914,7 +924,7 @@ def make_grid(grid_type, images, ra, dec, percentiles=[], spacing=0.75,
                                               size=size, spacing=spacing)
 
     if grid_type == "single":
-        ra_grid, dec_grid = [ra], [dec]
+        ra_grid, dec_grid = [sim_galra], [sim_galdec]
```
rknop, line 927: Raise an exception here if the two needed parameters are None.  (I suggest renaming them above to `singlera` and `singledec`.)

campari/AllASPFuncs.py
rknop: (Don't need to fix this for this PR, but it might be worth starting to file issues for things we know that need to be changed because they still have OU-specific code in them.)

campari/AllASPFuncs.py
```diff
@@ -1178,17 +1201,18 @@ def build_lightcurve(ID, exposures, confusion_metric, flux, sigma_flux, ra, dec)
...
-    band (str): the bandpass of the images used
+    flux (array): the output flux of the SMP algorithm. If no flux is received,
+                    then no lightcurve is built.
+    sigma_flux (array): the output flux error of the SMP algorithm
     ra, dec (float): the RA and DEC of the object.
 
     Returns:
     lc: a QTable containing the lightcurve data
     """
+    # If running without a detection, don't build a lightcurve.
```
rknop, line 1213: ...when does this happen?  This seems a bit weird to me to have in this function.  It would make more sense to insist that flux and sigma_flux not be None, and then have the thing that calls this just not call this function if there aren't any fluxes to build with.

campari/AllASPFuncs.py
```diff
@@ -1282,9 +1306,13 @@ def build_lightcurve_sim(supernova, flux, sigma_flux):
     Returns
     lc: a QTable containing the lightcurve data
     """
+    # If running without a detection, don't build a lightcurve.
```
rknop, line 1309: Same comment as in `build_lightcurve`

campari/AllASPFuncs.py
```diff
@@ -1307,6 +1335,9 @@ def save_lightcurve(lc, identifier, band, psftype, output_path=None,
     The file name is:
     <output_path>/identifier_band_psftype_lc.ecsv
     """
+    if lc is None:
```
rknop, line 1338: Same comment (effectively) as in `build_lightcurve` ; `lc=None` should be an error, and the thing that calls should just not call this function if it doesn't have a lc.

campari/simulation.py
```diff
@@ -220,13 +225,17 @@ def simulate_wcs(angle, x_shift, y_shift, roman_path, base_sca, base_pointing,
...
     if deltafcn_profile:
+        SNLogger.debug("Using delta function profile for galaxy.")
         profile = galsim.DeltaFunction()
     else:
-        bulge = galsim.Sersic(n=3, half_light_radius=1.6)
-        disk = galsim.Exponential(half_light_radius=5)
+        SNLogger.debug("Using bulge+disk profile for galaxy. The bulge has a half-light radius of "
+                          f"{bulge_hlr} and the disk has a half-light radius of {disk_hlr}.")
+        bulge = galsim.Sersic(n=3, half_light_radius=bulge_hlr)
+        disk = galsim.Exponential(half_light_radius=disk_hlr)
```
rknop, line 238: OK, I'm a bit confused.  If you're simulating a galaxy with a couple of profiles like this, then isn't using a single grid point at the position of the galaxy the wrong thing to do when trying to fit to the simulation?  You'd need multiple PFSes spread out to get the smooth profile of the galaxy.

Or am I misunderstanding what you do with the `sim_galra` and `sim_galdec` parameters to `make_grid`?

campari/AllASPFuncs.py
ColeFMeldorf: But we'd still want it to be a snappl PSF object right? So is it as simple as removing the "ou24PSF_slow" and replace it with a variable? Or is it more in - depth than that?

rknop: Yeah, it'd still be a snappl PSF.  You'd take the constructing of the PSF here out and make it a passed parameter in the function.

campari/AllASPFuncs.py
ColeFMeldorf: check!

campari/simulation.py
ColeFMeldorf: I think my code is making it a little confusing about what's going on, and this will also probably explain why I chose to name those other variables sim_galra and sim_galdec.

When I simulate an image, the galaxy is always placed at sim_galra and sim_galdec. It is not always true that we are fitting with a single grid point. If the grid method is "single" the goal is to check and make sure that grid points are being placed exactly where I intend them to. Thus, we simulate a galaxy that is a delta function, (at sim_galra and sim_galdec) and then try to place the grid point right on top of it, theoretically resulting in machine-precision recovery of the correct image. I'll add a warning that says something if grid type = single is passed alongside delta function = False, as you'd never want to do that.

With this information, should I change sim_galra and sim_galdec still? They are used even when grid =/= single!

campari/AllASPFuncs.py
ColeFMeldorf: See my comment in response to your first comment in simulation.py

campari/AllASPFuncs.py
ColeFMeldorf: Done!

campari/AllASPFuncs.py
ColeFMeldorf: Done!

campari/AllASPFuncs.py
ColeFMeldorf: Done!

PR #140, opened 2025-08-05
Simulation Grids (not WIP)

ColeFMeldorf:
In this PR we add the ability to pass a list of simulation parameters and create a grid. So for instance if I requested:
BG Gal Flux = [100, 500]
BG Gal Separation = [0.1, 0.5] 
Then the code should run a simulation of galaxies with (flux, separation) combinations of: (100, 0.1) (100, 0.5) (500, 0.1) (500, 0.5), e.g.
To do this I make the following changes:
1. [x] The galaxy separation and shape were each controlled by two parameters each, the ra and dec separation and the bulge and disk half light radii, respectively. To make running multiple simulations easier, I have reworked each into single variables, `sim_galaxy_scale `and `sim_galaxy_offset`
2. [x] Now when the galaxy separation, size, and flux are read in, they have the suffix _all on their variable name. This is so that I can loop over these lists in the grid, `for sim_galaxy_scale in sim_galaxy_scale_all` e.g.
3. [x] At lines 340-350 ish, I perform a meshgridding on all of the simulated parameters
4. [x] These simulation parameters, along with a run name, are now used to create the light curve name when saving for simulated runs.

Note that while there are now new variables for sim_galaxy_scale and sim_galaxy_separation, the old variables, the ra and dec sep and the bulge and disk hlr, still work.

### DIFF
--- a/changes/140.campari.rst
+++ b/changes/140.campari.rst
@@ -1,0 +1,1 @@
+Made grids of sim parameters available.


campari/RomanASP.py
```diff
@@ -236,26 +236,27 @@ def main():
...
-    bulge_hlr = config.value("photometry.campari.simulations.bulge_hlr")
-    disk_hlr = config.value("photometry.campari.simulations.disk_hlr")
+    run_name = config.value("photometry.campari.simulations.run_name")
 
     if grid_type == "single" and not deltafcn_profile:
         SNLogger.warning("Using a single point on the grid without a delta function profile is not recommended."
                          "The goal of using a single point is to run an exact fit for testing purposes, which requires"
                          "the galaxy be a delta function.")
 
+
```
rknop, line 258: Make an issue : this main() is getting really really long.  It might be good to break it out more into functions.

PR #145, opened 2025-08-11
campari_runner rough draft

ColeFMeldorf:


campari/AllASPFuncs.py
```diff
@@ -1209,12 +1207,11 @@ def build_lightcurve(ID, exposures, confusion_metric, flux, sigma_flux, ra, dec)
     Returns:
     lc: a QTable containing the lightcurve data
     """
-
-
     flux = np.atleast_1d(flux)
     sigma_flux = np.atleast_1d(sigma_flux)
     band = exposures["filter"][0]
     mag, magerr, zp = calc_mag_and_err(flux, sigma_flux, band)
+    SNLogger.debug(exposures)
```
rknop, line 1214: For debug strings, it's often worth putting in some text to explain.  If you're just watching the output as it goes by, it lets you know where things are if it's not cryptic.  Sometimes later you will search log files for specific strings.

Even for really long debug strings like this, you might add a bit of introduction, something like:

`SNLogger.debug( f"Building lightcurve for {exposures}" )`

Or, even, if you want to make it more readable,:
`SNLogger.debug( f"Building lightcurve for:\n{'\n'.join( [ f'  * {e}' for e in exposures ] )}" )`

campari/campari_runner.py
```diff
@@ -0,0 +1,381 @@
...
+    save_lightcurve,
+)
+
+
+class campari_lightcurve_model:
+    """This class holds the output of the Campari pipeline for a single SNID."""
+
+    def __init__(self, flux, sigma_flux, images, model_images, exposures, ra_grid, dec_grid, wgt_matrix,
+                 confusion_metric, best_fit_model_values, cutout_wcs_list, sim_lc):
+        """Initialize the Campari lightcurve model with the SNID and its properties."""
```
rknop, line 35: Add a Parmaeters section to the docstring that briefly defines what all the arguments are.

(Or, if you don't want to do it for this PR, create an issue telling somebody to do it.)

Also, see comment below where you actually make one of these.  Give everything other than self a default.  Either a sensible default, or just None.  That makes it clear that they this constructor is supposed to be called with kwargs, not with args.  (If you want to be really anal, you could throw an exception if any of the necessary ones are None.  But, you could also just count on the code crashing later if something is None that shouldn't be.)

campari/campari_runner.py
```diff
@@ -0,0 +1,381 @@
...
+
+
+class campari_runner:
+    """This class is used to run the Campari pipeline."""
+
+    def __init__(self, **kwargs):
+        """Initialize the Campari runner with all of the variables needed to run the pipeline."""
+
+        try:
+            self.cfg = Config.get(kwargs["config"], setdefault=True)
```
rknop, line 57: Hmm, OK.  `setdefault=True` is very scary for config.  (If you look at the `Config.get` docstring, it says "Avoid use of this" at the beginning of the description of that argument.)  I wouldn't do it deep inside this function here; that's something I would only do in main when you can be very sure that this is the very first time you're initializing the config during the running of the program.

The problem is that if you write something somewhere that uses this class from elsewhere in python, changing the config here pulls the rug out from underneath the rest of the code.  (I speak from experience; I did exactly this in a different pipeline, and it raised problems down the road.)

Here's what I would do:

In fact, looking at it, main() already completely handles the "config" argument.  So, here, you should never be passing an argument to `Config.get`, nor should you ever use `setdefault` with `Config.get`.  Just do `self.cfg = Config.get()`.  In that case, there's no need to put it inside a `try/except` block.

In this function, totally ignore the "config" argument that was passed in.  Perhaps document in the docstring that this will happen, and that config is expected to be fully set before this function is called (perhaps with a reference to your main as a place where that happens).

campari/campari_runner.py
```diff
@@ -0,0 +1,381 @@
...
+        self.healpix_file = kwargs["healpix_file"]
+        self.nside = kwargs["nside"]
+        self.object_lookup = kwargs["object_lookup"]
+        self.transient_start = kwargs["transient_start"]
+        self.transient_end = kwargs["transient_end"]
+
+        self.ra = kwargs["ra"]
+        self.dec = kwargs["dec"]
+
+        config = Config.get(kwargs["config"], setdefault=True)
```
rknop, line 85: Omit this line, and everywhere below you do `config.value`, do `self.cfg.value`.

campari/campari_runner.py
```diff
@@ -0,0 +1,381 @@
...
+                                                            np.sort(self.pointing_list)):
+            SNLogger.warning("Unable to find the object in all the pointings in the image list. Specifically, the"
+                             " following pointings were not found: "
+                             f"{np.setdiff1d(self.pointing_list, exposures['pointing'])}")
+
+        return exposures
+
+    def get_sedlist(self, ID, exposures):
+        """Create a list of SEDs for the given SNID and exposures."""
+        sed_obj = OU2024_Truth_SED(ID, isstar=(self.object_type == "star")) if self.fetch_SED else Flat_SED()
```
rknop, line 302: Make an issue somewhere (if there isn't one that covers this already) that this is something else that will need to be addressed to make the code not OU24 specific.

campari/campari_runner.py
```diff
@@ -0,0 +1,381 @@
...
+    def call_run_one_object(self, ID, ra, dec, exposures, sedlist, param_grid_row):
+        """Call the run_one_object function to run the pipeline for a given SNID and exposures."""
+        if not self.use_real_images:
+            bg_gal_flux, sim_galaxy_scale, sim_galaxy_offset = param_grid_row
+        else:
+            bg_gal_flux, sim_galaxy_scale, sim_galaxy_offset = None, None, None
+
+        flux, sigma_flux, images, sumimages, exposures, ra_grid, dec_grid, wgt_matrix, \
+            confusion_metric, X, cutout_wcs_list, sim_lc = \
+            run_one_object(ID, ra, dec, self.object_type, exposures, self.roman_path,
```
rknop, line 317: If you have more than four or five arguments, it's very dangerous and error-prone to just list all the arguments in order.  It's very easy to omit one and have everything be wrong, and it's a huge pain in the butt then to figure out what happened.  I would strongly recommend rewriting this call to have a series of "param=value" lines.  Not only does that make it safer for you, it makes it much easier to fix things if you ever need to change what is passed to `run_one_object`, and it makes it a lot more readable.

I would also edit the definition of `run_one_object` in `AllASPFuncs.py` to have `=None` as a default for all the arguments there, to make it clear they're expected to be called as kwargs, not args.  (Or, you could put in sensible defaults.  But, if there's no sensible default, hopefully putting in None will cause the code to crash somewhere, so if a necessary argument is omitted the code won't seem to run right.)

campari/campari_runner.py
```diff
@@ -0,0 +1,381 @@
...
+                           self.sn_path, self.size, self.band, self.fetch_SED, sedlist, self.use_real_images,
+                           self.use_roman, self.subtract_background,
+                           self.make_initial_guess, self.initial_flux_guess,
+                           self.weighting, self.method, self.grid_type,
+                           self.pixel, self.source_phot_ops, self.do_xshift, bg_gal_flux,
+                           self.do_rotation, self.airy, self.mismatch_seds, self.deltafcn_profile,
+                           self.noise, self.check_perfection, self.avoid_non_linearity,
+                           self.spacing, self.percentiles, sim_galaxy_scale,
+                           sim_galaxy_offset, base_pointing=self.base_pointing, base_sca=self.base_sca)
+        lightcurve_model = campari_lightcurve_model(
```
rknop, line 327: Same comment as on run_one_object.

campari/tests/test_campari.py
```diff
@@ -98,12 +97,24 @@ def test_find_all_exposures(roman_path):
...
                                  truth="simple_model")
-    compare_table = ascii.read(pathlib.Path(__file__).parent
-                               / "testdata/findallexposurestest.dat")
-    assert explist["pointing"].all() == compare_table["pointing"].all()
-    assert explist["sca"].all() == compare_table["sca"].all()
 
-    assert explist["date"].all() == compare_table["date"].all()
+    np.save("./testdata/findallexposures.npy", explist)
+
+    compare_table = np.load(pathlib.Path(__file__).parent / "testdata/findallexposures.npy")
```
rknop, line 103: This test looks completely circular to me.

* Write `explist` to `testdata/findallexposures.npy`
* Make sure that `explist` has the same contents as `testdata/findallexposures.npy`

Unless I've missed something, this is a test of `np.save/np.load`, not of your code.

campari/tests/test_runner.py
```diff
@@ -0,0 +1,303 @@
...
+def test_get_exposures(cfg):
+    test_args = create_default_test_args(cfg)
+    test_args.object_lookup = True
+    test_args.SNID = 20172782
+
+    runner = campari_runner(**vars(test_args))
+    runner.decide_run_mode()
+    explist = runner.get_exposures(7.731890048839705, -44.4589649005717, 62654.0, 62958.0)
+
+    np.save("./testdata/findallexposures.npy", explist)
```
rknop, line 213: Another test that's circular : write a file, make sure the file contents are what you just wrote to it.

campari/tests/test_runner.py
```diff
@@ -0,0 +1,303 @@
...
+    exposures = pd.DataFrame(data={"date": [1, 2, 3], "filter": ["Y106", "Y106", "Y106"],
+                                   "detected": [True, True, True],
+                                   "pointing": [1, 1, 1], "sca": [1, 1, 1], "x": [0, 0, 0], "y": [0, 0, 0],
+                                   "x_cutout": [0, 0, 0], "y_cutout": [0, 0, 0]})
+    exposures = Table.from_pandas(exposures)
+    ra_grid = np.array([1, 2, 3])
+    dec_grid = np.array([1, 2, 3])
+    wgt_matrix = None
+    confusion_metric = None
+    best_fit_model_values = np.array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0,  0, 0, 0, 0, 0, 0])
```
rknop, line 262:
```suggestion
    best_fit_model_values = np.array([0] *16, dtype=float)
```

campari/tests/test_runner.py
```diff
@@ -0,0 +1,303 @@
...
+    cutout_wcs_list = None
+    sim_lc = None
+    ra = 7.731890048839705
+    dec = -44.4589649005717
+
+    lc_model = campari_lightcurve_model(flux, sigma_flux, images, model_images,
```
rknop, line 268: As commented above, I recommend changing this from a sequence of arguments in order to a series of `key=value` parameters.

campari/RomanASP.py
```diff
@@ -82,9 +63,9 @@ def main():
         #   may mean that things will fail later, but it may also just mean
         #   that somebody is doing "--help"
         cfg = None
-        desc += (" Include --config <configfile> before --help "
-                 "(or set SNPIT_CONFIG) for help to show you all config "
-                 "options that can be passed on the command line.")
+    desc += (" Include --config <configfile> before --help "
```
rknop, line 66: MWV might disagree with this, but this is what I've done with phrosty.

I would replace *everything* inside this except block (including the comment) with :
```
    except RuntimeError as e:
        if str(e) == 'No default config defined yet; run Config.init(configfile)':
            sys.stderr.write( "Error, no configuration file defined.\n"
                              "Either run campari with -c <configfile>\n"
                              "or set the SNPIT_CONFIG environment varaible.\n" )
            sys.exit(1)
        else:
            raise
```

Reasoning: this gives you enough information to get the help you're actually looking for on the next command.  If you allow it to give help without having a valid config, your help string will omit a whole bunch of the things it would normally include, so the help information isn't actually right.  Plus, I'm very nervous about going forward at this point without a config; it probably requires you to do the "setdefault" stuff you did in the campari arg parse that is really a bad idea.

Yes, it's a little cumbersome to have to specify a config file when asking for help.  But there's a fundamental issue that you can't have arguments automatically added based on what's in a config file without having a config file.  This speedbump on getting basic help is a reasonable tradeoff.

campari/campari_runner.py
```diff
@@ -0,0 +1,381 @@
...
+        self.best_fit_model_values = best_fit_model_values
+        self.cutout_wcs_list = cutout_wcs_list
+        self.sim_lc = sim_lc
+
+
+class campari_runner:
+    """This class is used to run the Campari pipeline."""
+
+    def __init__(self, **kwargs):
+        """Initialize the Campari runner with all of the variables needed to run the pipeline."""
```
rknop, line 54: Write a "Parameters" section that briefly documents all the arguments that can be set here.  (You can cut and paste from your argparse help strings.)

rknop: (No need to document all the arguments in the config file; eventually there should be a sample or default config that has comments that document all of those.   Just document the things you're reading directly out of `kwargs`.  As I note below, assuming you make the changes to config handling, note here that `config` is a legal argument for compatibility with your main, but it will be totally ignored by this constructor.)

campari/campari_runner.py
```diff
@@ -0,0 +1,381 @@
...
+
+        if lc_model.flux is not None:
+            output_dir = pathlib.Path(self.cfg.value("photometry.campari.paths.output_dir"))
+            save_lightcurve(lc, identifier, self.band, psftype, output_path=output_dir)
+
+        # Now, save the images
+        images_and_model = np.array([lc_model.images, lc_model.model_images, lc_model.wgt_matrix])
+        debug_dir = pathlib.Path(self.cfg.value("photometry.campari.paths.debug_dir"))
+        SNLogger.info(f"Saving images to {debug_dir}")
+        np.save(debug_dir / f"{identifier}_{self.band}_{psftype}_images.npy", images_and_model)
```
rknop, line 365: Not for now, but an issue: eventually you're going to want to have an option to turn off all the debug saving.

campari/AllASPFuncs.py
ColeFMeldorf: I actually meant to remove this, whoops!

campari/campari_runner.py
ColeFMeldorf: Done!

campari/campari_runner.py
ColeFMeldorf: Done and done!

campari/campari_runner.py
ColeFMeldorf: issue made!

campari/campari_runner.py
ColeFMeldorf: Done!

rknop (approved):

PR #150, opened 2025-08-27
Saving Galaxy and Noise of simulated images.

ColeFMeldorf:
Campari now has the ability to save:
1. [x] The noise that is added to a simulated image
2. [x] The galaxy that was used in a simulated image
3. [x] The model prediction image for _just_ the galaxy. I.e. the model - supernova.

These will be useful for me for debugging. I am trying to determine where exactly the algorithm is becoming nudged off of the correct flux.

### DIFF
--- a/examples/perlmutter/campari_config_jupyter.yaml
+++ b/examples/perlmutter/campari_config_jupyter.yaml
@@ -17,10 +17,12 @@ photometry:
     subtract_background: true
     use_real_images: true
     save_model: true
+    save_debug: true
 
     paths:
       sims_sed_library: /pscratch/sd/c/cmeldorf/rubin_sim_data/sims_sed_library
       roman_path: /dvs_ro/cfs/cdirs/lsst/shared/external/roman-desc-sims/Roman_data
+      sims_dir: /dvs_ro/cfs/cdirs/lsst/shared/external/roman-desc-sims/Roman_data
       sn_path: /dvs_ro/cfs/cdirs/lsst/www/DESC_TD_PUBLIC/Roman+DESC/PQ+HDF5_ROMAN+LSST_LARGE
       output_dir: /pscratch/sd/c/cmeldorf/campari_out_dir
       debug_dir: /pscratch/sd/c/cmeldorf/campari_debug_dir
--- a/campari/RomanASP.py
+++ b/campari/RomanASP.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 # Standard Library
 import argparse
 import warnings
--- a/campari/campari_runner.py
+++ b/campari/campari_runner.py
@@ -141,6 +141,7 @@ class campari_runner:
         self.base_pointing = self.cfg.value("photometry.campari.simulations.base_pointing")
         self.base_sca = self.cfg.value("photometry.campari.simulations.base_sca")
         self.run_name = self.cfg.value("photometry.campari.simulations.run_name")
+        self.save_debug = self.cfg.value("photometry.campari.save_debug")
         self.param_grid = None
         self.run_mode = None
         self.noise_maps = None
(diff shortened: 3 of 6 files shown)


campari/AllASPFuncs.py
```diff
@@ -1885,9 +1890,9 @@ def run_one_object(ID=None, ra=None, dec=None, object_type=None, exposures=None,
         # possible. In the meantime, just return zeros for the simulated lc
         # if we aren't simulating.
         sim_lc = np.zeros(num_detect_images)
-    return flux, sigma_flux, images, sumimages, exposures, ra_grid, dec_grid, \
+    return flux, sigma_flux, images, model_images, galaxy_only_model_images, exposures, ra_grid, dec_grid, \
```
rknop, line 1893: Don't change this here, but make an issue to make this better.

In general, if you're returning more than 2 or 3 things in a tuple from a function, it becomes extremely error-prone.  On the other end, did you get all the variables in the right order?  When you change (as you do in this PR), did you change everywhere properly?

There are a couple of things you can do.  You can return a dict, so it's clear what is what.  However, my preference is that this indicates a situation where the thing should be refactored to an object-oriented design.  This is crying out for some sort of SimulatedThingy class, or some such, that has the properties that are all the things that you're returning here.

campari/simulation.py
```diff
@@ -107,6 +107,8 @@ def simulate_images(num_total_images, num_detect_images, ra, dec,
     sn_storage = []
     image_list = []
     cutout_image_list = []
+    noise_maps = []
+    galaxy_images = []
```
rknop, line 111: Making this comment here and not on the function call, because the interface doesn't give me an option to do it there:

File an issue to make most of the things you pass to the function call kwargs rather than args; do this by making them defaults.  If they are required, you can default them to None and put a check at the begining of the function that they are None.

Reason: having a large number of positional parameters is error prone and makes the code very hard to read.  When debugging a function call, you have to look at two lists at once, and make sure that all the things are all in the right order.  If everything has to be a kwarg, then you have a bunch of `key=value` in the function call, and that's a whole lot easier to read and interpret.

campari/simulation.py
```diff
@@ -216,7 +223,7 @@ def simulate_images(num_total_images, num_detect_images, ra, dec,
     SNLogger.debug(f"images shape: {images[0].shape}")
     SNLogger.debug(f"images length {len(images)}")
 
-    return sim_lc, util_ref, image_list, cutout_image_list, galra, galdec
+    return sim_lc, util_ref, image_list, cutout_image_list, galra, galdec, galaxy_images, noise_maps
```
rknop, line 226: Same comment as earlier re: making an issue re: large number of return values.

I really think you ought to think about having some sort of simulated thingy class, and make the simulations more object oriented.

(There is a danger in doing that of eating up memory with no tomorrow, though.)

examples/perlmutter/campari_config_jupyter.yaml
```diff
@@ -1,5 +1,5 @@
 ou24:
-  config_file: /pscratch/sd/c/cmeldorf/campari/examples/perlmutter/tds.yaml
+  config_file: /pscratch/sd/c/cmeldorf/campari/examples/perlmutter/tds_jupyter.yaml
```
rknop, line 2: Just as a note : examples that point into your private scratch space may not work for anybody else.

My preference is to run things in the docker image, have paths in the config file be inside the docker image, and then tell people to bind-mount the right places.

Indeed, right now, this example is really only useful for you.  It might accidentally be useful for somebody else if they are able to read your file.  But if you really want an example to be a usable example, even if it's designed for a specific system (such as Perlmutter), you should package in the archive everything that's needed for the example to work.  (Within reason; don't package all the example files, but you can tell people they need the `photometry_test_data` archive or some such.)

See https://github.com/Roman-Supernova-PIT/phrosty/tree/main/examples/perlmutter as an example of how this kind of example might work.  Notice there that the `.yaml` file points to things inside the container.  The `.sh` file that runs the container has stuff relative to `$PWD`, and the instructions for running the example in the documentation tell people what directories they need to create.  Nowhere does it depend on files existing on the system in any of mine or Lauren's directories.  (It _does_ depend, however, on the ou2024 sims being in the standard place, as they are mounted there in the `interactive_podman.sh` script.  THis means eventually this example will break.  But, it's better to point to standard data repository directories than anything that's in anybody's user directory anywhere.)

Don't try to fix this for this PR, but make an issue, and keep this in mind as a general thing.

(Also, I hate jupyter... I can see the value in using them for examples,and I often use them for documentation, but if you really want these examples to help people run this for real, examples that run outside of jupyter are good to have too.)

rknop (commented):
As a general note : I won't review juptyer scripts when doing code reviews.  Nothing essential should ever be in a jupyter script for repositories like this.

campari/simulation.py
ColeFMeldorf: Makes sense!

campari/simulation.py
ColeFMeldorf: Got it!

PR #151, opened 2025-08-29
Implement snappl `ImageCollection`

ColeFMeldorf:
In this PR we make use of the new snappl ImageCollection class. This mainly changes the `find_all_exposures` function.

1. [x] Currently, to load images, `find_all_exposures` would search for images and create an exposures object, which was then used by construct_images to build the cutouts. Now the ImageCollection directly returns a list of snappl Image objects. This makes the exposures object partially redundant, but not entirely. For now, I pass around both `exposures`, a table of information about the images,  and `image_list`, the latter being the output of ImageCollect.find_images. In the future, it might be possible to just pass around the image list, and turn it into a table of information at the very end when saving lightcurves, which is essentially what I do in lines 417-435. 
2. [x] ImageCollection expects `None` for no MJD limit so I have replaced all of my np.inf with None.


campari/AllASPFuncs.py
```diff
@@ -384,71 +385,67 @@ def find_all_exposures(ra, dec, transient_start, transient_end, band, maxbg=None
...
     if maxdet is not None:
-        det = det.iloc[:maxdet]
-    det["detected"] = True
+        transient_images = transient_images[:maxdet]
 
-    if pointing_list is not None:
-        det = det.loc[det["pointing"].isin(pointing_list)]
-    bg = res.loc[(res["date"] < transient_start) | (res["date"] > transient_end)].copy()
-    bg = bg.loc[(bg["date"] >= image_selection_start) & (bg["date"] <= image_selection_end)]
+    transient_sca = [img.sca for img in transient_images]
```
rknop, line 418: Rather than keep variables `transient_sca`, `transient_pointing`, and `transient_mjd`, you should just use `transient_images`, and then pull out the properties of those objects as you need them later in the function.  (Right now, you have duplicated data, as the same information in stored in two places.  The only time you should really do that is if you need to modify one and you don't want to modify the other, or if it's really necessary for clarity or brevity.  I don't think either is the case here.)

campari/AllASPFuncs.py
```diff
@@ -384,71 +385,67 @@ def find_all_exposures(ra, dec, transient_start, transient_end, band, maxbg=None
...
+    transient_pointing = [img.pointing for img in transient_images]
+    transient_mjd = [img.mjd for img in transient_images]
+
+    no_transient_sca = [img.sca for img in no_transient_images]
```
rknop, line 422: Same comment for these three variables as for the three `transient_*` variables.

campari/AllASPFuncs.py
```diff
@@ -354,7 +355,8 @@ def construct_static_scene(ra, dec, sca_wcs, x_loc, y_loc, stampsize, psf=None,
 def find_all_exposures(ra, dec, transient_start, transient_end, band, maxbg=None,
                        maxdet=None, return_list=False,
                        roman_path=None, pointing_list=None, sca_list=None,
-                       truth="simple_model", image_selection_start=-np.inf, image_selection_end=np.inf):
+                       truth="simple_model", image_selection_start=None, image_selection_end=None,
+                       image_source="ou2024"):
```
rknop, line 359: Remove the docstring sentence about Rob's awesome database method, since that's no longer used in this function.  (That's all moved to snappl.)

campari/AllASPFuncs.py
rknop: ....in fact, there's a much bigger comment I'm going to leave on this whole PR that addresses all of this.

rknop (commented):
I only read the beginning of this, because there's a lot more that should be done here.

Once you're using ImageCollection, you no longer need to keep an astropy table or a Pandas dataframe of all the exposure info.  Just keep the list of Image objects that you got back from the ImageCollection, and then access the fields of those objects.  The whole point of the snappl `Image` class is that it encapsulates the information you need about an image, so you shouldn't need to maintain a set of tables along side a list of Images that has all that same information.

This will have changes in lots of places -- starting in `find_exposures` which can get a whole lot shorter and should just return a list of the `Image` objects that you found (which, I believe, will be a concatenation of `no_transient_images` and `transient_images`).

The find everything that calls `find_exposures`, or calls something that calls it, and refactor it to use that list of images rather than all the pandas dataframes and such that you built.

ColeFMeldorf:
Ok I've refactored to get rid of all instances of "exposures", the pandas table containing all of the info about the images with image_list, a list of snappl image objects. All tests are passing but the changes are a bit thorny so here are some summaries:

- Any function that takes in exposures as an argument now takes in image_list as an argument (and sometimes diaobj gets added as well)
- These functions' tests have been changed.
-Because of there no longer being an exposures object, there is no good way to pass around x and y cutout coordinates with the images. Therefore when building lightcurves at the end,  I manually recalculate x and y using the ra, dec, and WCS. Thus, some tests had to change their reference values because I could no longer manually enter a dummy value for x and y.
-Any time there was a list of parameters derived from the exposures table, there is now a list that gets unpacked from image_list using a list comprehension. For instance, if I want the pointings I do `[a.pointing for a in image_list]` .
-With all the things moving to snappl, `fetch_images ` is now redundant and has been removed.

PR #158, opened 2025-09-03
Allowing campari to handle partial overlaps

ColeFMeldorf:
When trying to run objects to test pipeline capabilities for Ben, I started getting issues where some cutouts only partially overlapped with the SCA. Campari can now handle these cases without throwing an error, pending a companion PR to snappl.


campari/campari_runner.py
```diff
@@ -187,6 +187,7 @@ def __call__(self):
             banner(f"Running SN {ID}")
 
             if self.object_lookup:
```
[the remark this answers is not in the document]
ColeFMeldorf: Yes, this is done in PR 153, (https://github.com/Roman-Supernova-PIT/campari/pull/153) but I've added the object_collection argument. I've added --oid as an alias for --SNID so that old code does not break.

campari/campari_runner.py
```diff
@@ -184,7 +189,8 @@ def __call__(self):
         for index, ID in enumerate(self.SNID):
             banner(f"Running SN {ID}")
 
-            if self.object_lookup:
+            if self.object_collection != "manual":
+                SNLogger.debug(f"Looking up object info for SN {ID}")
                 ra, dec, transient_start, transient_end = self.lookup_object_info(ID)
             else:
                 ra, dec, transient_start, transient_end = self.ra, self.dec, self.transient_start, self.transient_end
```
rknop, line 196: I don't think I fully communicated what I was after before.  You should not have any code yourself here that behaves differently based on the value of `manual_collection`.  If you need that, that means that `snappl` needs more stuff.  You should just pass that on to `snappl`, and then just use the `DiaObject` you getback.

```suggestion
            SNLogger.debug(f"Looking up object info for SN {ID}")
            diaobj = DiaObject.find_objects( collection=self.object_collection, id=ID, ra=self.ra, dec=self.dec )
            # Put in a check here that you got back only one object, yell if you got back 0 or more
            diaobj = diobj[0]
            if self.ra is not None:
                if np.fabs( self.ra - diaobj.ra ) > 1. / 3600. / np.cos( dioaobj.dec * np.pi / 180. ):
                    SNLogger.warning( f"Given RA {self.ra} is far from DiaObject nominal RA {diaobj.ra}" )
                diaobj.ra = self.ra
            if self.dec is not None:
                if np.fabs( self.dec - diaobj.dec ) > 1. / 3600.:
                    SNLogger.warning( f"Given Dec {self.dec} is far from DiaObject nominal Dec {diaobj.dec}" )
                diaobj.dec = self.dec
            if self.transient_start is not None:
                # Maybe check if it's different from what's in DiaObject, if the thing in DiaObject isn't None?
                diaobj.mjd_start = self.transient_start
            if self.transient_end is not None:
                diaobj.transient_end = self.transient_end
```

Further changes you should do: don't make ra, dec, transient_start, or transient_end variables in self any more.  Instead, make a `self.diaobj`, and set the `diaobj` I have above to that.  Then, when you need `ra`, `dec`, `mjd_start`, or `mjd_end`, pull them out of `self.diaobj`.

rknop: ....OK, this is maybe too much for this.  I'm going to make an issue, because there are other bigger issues.  So, don't put in this change yet.

rknop (approved):

campari/campari_runner.py
ColeFMeldorf: Sounds good. I'll just merge this for now and address that  issue.

PR #165, opened 2025-09-16
A few small fixes that prevent hangups when parsing input files

ColeFMeldorf:


campari/campari_runner.py
```diff
@@ -186,7 +187,11 @@ def decide_run_mode(self):
 
         # Option 1, user passes a file of SNIDs
         if self.SNID_file is not None:
-            self.SNID = pd.read_csv(self.SNID_file, header=None).values.flatten().tolist()
+            if ".dat" not in str(self.SNID_file) and ".DAT" not in str(self.SNID_file):
+                self.SNID = pd.read_csv(self.SNID_file, header=None, comment="#").values.flatten().tolist()
+            else:
+                self.SNID = list(yaml.safe_load(open(self.SNID_file, "r")).values())[0]
+
```
rknop, line 194: I'm very confused about all of this.

What is in the SNID file?  Why have both yaml and ... csv?  Elsewhere in the code, it looks like `self.SNID` is just a list of IDs, which would be a list of integers.  If so, then all of this is tremendously overdone.  You REALLY don't need to have calls to heavyweight libraries like pandas to read a text file of integers.  You don't need to read CSV files, you don't need to read YAML files.  Just support a plain text file that has one ID per line, and read that without using any csv, yaml, or pandas libraries.

That makes this next comment irrelevant, but: multiple problems with your filename detection.  It seems that EVERYTHING that doesn't have `.dat' or '.DAT' in it is a CSV file.  It's perverse enough to consider things that don't end in .dat to be a CSV file (so, image.fits is a CSV file...), but also `file.dat.csv` would here be detected as not a CSV file.  If you want to check the end of the filename, check the end of the filename.  Next perverse thing: if and only if the filename has `.dat` in it,it's a ... yaml file?????

I don't get any of this.

What are you really trying to do here?  If you want a way of passing a whole bunch of SNIDs without having to list them all on the command line, then just *define* that the `SNID_file` has one integer per line and do

```
self.SNID = []
with open( self.SNID_file ) as ifp():
    for line in ifp:
        if len( line.strip() ) > 0:
            self.SNID.append( int(line) )
```

or something like that.  Don't think about yaml, don't think about csv, unless you really do need tabular data with lots of columns (csv) or if you do need a dictionary or some such type structure (yaml).

ColeFMeldorf:
Addressed comments!

campari/campari_runner.py
```diff
@@ -186,7 +186,12 @@ def decide_run_mode(self):
 
         # Option 1, user passes a file of SNIDs
         if self.SNID_file is not None:
-            self.SNID = pd.read_csv(self.SNID_file, header=None).values.flatten().tolist()
+            self.SNID = []
+            with open(self.SNID_file, "r") as ifp:
+                for line in ifp:
+                    # Check that the line is non-empty, not a comment, and doesn't contain letters.
+                    if len(line.strip()) > 0 and line[0] != "#" and not any(c.isalpha() for c in line):
```
rknop, line 193: Do you want to skip letters, or do you want to error out if there are letters?  Right now it will just skip anything with letters, but it might be better to error out, because that means something's wrong with the file.  Plus, if what you need is an integer, there are lots of thing other than letters that are bad.

I would also do: `line.strip()[0] != '#'`, so that lines that start with spaces and then a `#` get skipped.

So, maybe:

```
line = line.strip()
if ( len(line) > 0 ) and ( line[0] != '#' ) and all( c in '0123456789' for c in line ):
```

if what you *really* want is to silently ignore lines with illegal characters, rather than have that be an error.  But, I would recommend making invalid lines an error rather than something to be skipped.  In that case, just let it thrown an exception when you do `int(line)` below, so  you end up with:
```

line = line.strip()
if ( len(line) > 0 ) and ( line[0] != '#' ):
```

PR #172, opened 2025-09-22
Changing defs in construct transient scene documentation to match snappl.

ColeFMeldorf:
The variable names and documentation in construct_transient scene vary from snappl in a way that was confusing.

To fix this I:

- [x] Renamed variables in construct_transient_scene to match snappl and changed their documentation and updated these variables wherever the function is called.
- [x] In construct_transient_scene I removed the checks on arguments because in other PRs @wmwv advised me to just let these things fail for a more informative error readout.
- [x] There were some small merge conflict issues that caused tests to fail. I had to move some small chunks of code around, add change .header --> set_fits_header, and restore the LSB argument to campari_lightcurve_model to ensure all tests would run. I am not sure how these got deleted in merges but I do have a lot of PRs open right now.

All tests pass.

### DIFF
--- a/changes/172.campari.rst
+++ b/changes/172.campari.rst
@@ -1,0 +1,1 @@
+Fixed construct transient scene documentation


wmwv (approved):
LGTM

PR #176, opened 2025-09-25
World's smallest fix!

ColeFMeldorf:
If passing an image list, don't check if there are pre and post detection images. We assume the user knows what they're doing.
This was causing a test to fail.


wmwv (commented):
LGTM

wmwv (approved):

PR #177, opened 2025-09-26
Removing Manual Fits Image to match new snappl paradigm

ColeFMeldorf:
`ManualFITSImage` is no more, so we update to `FITSImageStdHeaders` objects. These are largely the same except:

- [x] All new instances of `FITSImageStdHeaders` require a path. My old ManualFITS did not have nor need paths. Since it is required to be a string, I pass "none".
- [x] Some parameters like mjd and band can no longer be set at init and instead have to be manually set.
- [x] When setting a WCS header, mjd, band, etc. need to be manually set.

### DIFF
--- a/campari/campari_runner.py
+++ b/campari/campari_runner.py
@@ -302,7 +302,7 @@ class campari_runner:
                 # cutout sized and the 4088 is used nowhere.
                 img = FITSImageStdHeaders(
                     header=None, data=np.zeros((4088, 4088)), noise=np.zeros((4088, 4088)),
-                    flags=np.zeros((4088, 4088)), path="none"
+                    flags=np.zeros((4088, 4088)), path="/dev/null"
                 )
                 img.mjd = faux_dates[i]
                 img.band = self.band
--- a/campari/campari_runner.py
+++ b/campari/campari_runner.py
@@ -10,7 +10,7 @@ import galsim
 
 # SN-PIT
 from snappl.diaobject import DiaObject
-from snappl.image import ManualFITSImage
+from snappl.image import FITSImageStdHeaders
 from snappl.sed import Flat_SED, OU2024_Truth_SED
 from snpit_utils.config import Config
 from snpit_utils.logger import SNLogger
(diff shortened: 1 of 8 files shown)


campari/campari_runner.py
```diff
@@ -300,12 +300,15 @@ def get_exposures(self, diaobj):
                 # These data sizes are arbitary. I just need a data array present in order to perform the cutout,
                 # otherwise snappl throws an error. No data is actually placed into these images until they are
                 # cutout sized and the 4088 is used nowhere.
-                img = ManualFITSImage(
+                img = FITSImageStdHeaders(
                     header=None, data=np.zeros((4088, 4088)), noise=np.zeros((4088, 4088)),
-                    flags=np.zeros((4088, 4088)), pointing=self.base_pointing, sca=self.base_sca, band=self.band,
-                    mjd=faux_dates[i]
+                    flags=np.zeros((4088, 4088)), path="none"
```
rknop, line 305: ....that's a lot of memory to allocate to not use.  I need to dig into the code to see how these are used, but the comment you have here suggest that there's a design problem somewhere, either in campari, snappl, or both.  Don't deal with this PR, because the issue was already there, but make a campari Issue that point to this line and says that you're allocating 382MB of memory that is never used, and that there must be a better way.

rknop: Instead of `path="none"`, do `path="/dev/null"`.  That's the general Unix thing to use for "I don't really want to read or write this file."

But.... this suggests that the basic snappl `Image` class should support the case where there is no Path.  Please make a `snappl` issue that asks for `Image` to support a `None` path.

campari/tests/test_campari.py
```diff
@@ -340,7 +340,7 @@ def test_make_regular_grid():
     test_dec = np.array([-44.263969, -44.263897, -44.263825, -44.263918, -44.263846,
                          -44.263774, -44.263868, -44.263796, -44.263724])
     for wcs in [snappl.wcs.AstropyWCS.from_header(wcs_dict)]:
-        img = ManualFITSImage(header=wcs_dict, data=np.zeros((25, 25)))
+        img = FITSImageStdHeaders(header=wcs_dict, path="none", data=np.zeros((25, 25)))
```
rknop, line 343: Again, do `path="/dev/null"`.

campari/tests/test_campari.py
```diff
@@ -362,7 +362,7 @@ def test_make_adaptive_grid():
                                  / "testdata/images.npy")
         SNLogger.debug(f"compare_images shape: {compare_images.shape}")
         image = compare_images[0].reshape(11, 11)
-        img_obj = ManualFITSImage(header=wcs_dict, data=image)
+        img_obj = FITSImageStdHeaders(header=wcs_dict, data=image, path="none")
```
rknop, line 365: `/dev/null`

campari/tests/test_campari.py
```diff
@@ -388,7 +388,7 @@ def test_make_contour_grid():
         compare_images = np.load(pathlib.Path(__file__).parent
                                  / "testdata/images.npy")
         image = compare_images[0].reshape(11, 11)
-        img_obj = ManualFITSImage(header=wcs_dict, data=image)
+        img_obj = FITSImageStdHeaders(header=wcs_dict, data=image, path="none")
```
rknop, line 391: `/dev/null`

campari/tests/test_campari.py
```diff
@@ -565,8 +565,12 @@ def test_build_lc():
...
 
     for i in range(len(explist["date"])):
-        img = ManualFITSImage(
-            header=None, data=np.zeros((4085, 4085)), noise=np.zeros((4085, 4085)), flags=np.zeros((4085, 4085)),
+        img = FITSImageStdHeaders(
+            header=None,
+            data=np.zeros((4085, 4085)),
+            noise=np.zeros((4085, 4085)),
+            flags=np.zeros((4085, 4085)),
+            path="none"
```
rknop, line 573: `/dev/null`.

In the issue about allocating 382MB of memory you never use, reference this line as well.

PR #179, opened 2025-09-27
In order to get Gaussian PSF tests working, added the ability to save and load static scene and transient scene models.

ColeFMeldorf:
As it turns out, the Gaussian PSFs are quite slow because each pixel requires an expensive 2D integral :((.
However I realized I can make this a bit more feasible if the diagnostic tests create 1 background model that is then reused for each different test we want to do. Note that this branch is reliant on 2 currently unmerged PRs, so 175 and 177 should be looked at first.


campari/tests/test_campari.py
```diff
@@ -183,16 +183,31 @@ def test_regression_function(campari_test_data):
...
+        "--photometry-campari-cutout_size",
+        "19",
+        "--photometry-campari-weighting",
+        "--photometry-campari-subtract_background",
+        "--no-photometry-campari-source_phot_ops",
+        "--prebuilt_static_model",
+        str(pathlib.Path(__file__).parent / "testdata/reg_psf_matrix.npy"),
+        "--prebuilt_transient_model",
+        str(pathlib.Path(__file__).parent / "testdata/reg_sn_matrix.npy")
+    ]
```
rknop: AK.  This is why I hate linters that autoreformat.  They obey various formatting rules strictly, even when it totally kills readability.

The previous way was way more readable because it had the values of arguments right after the arguments on the same line.

campari/run_one_object.py
```diff
@@ -65,7 +65,8 @@ def run_one_object(diaobj=None, object_type=None, image_list=None, size=None, ba
                    airy=None, mismatch_seds=None, deltafcn_profile=None, noise=None,
                    avoid_non_linearity=None, spacing=None, percentiles=None, sim_galaxy_scale=1,
                    sim_galaxy_offset=None, base_pointing=662, base_sca=11,
-                   draw_method_for_non_roman_psf="no_pixel"):
+                   draw_method_for_non_roman_psf="no_pixel", save_model=False, prebuilt_psf_matrix=None,
```
rknop: ...this scares me.  I know it was there before, but options like this suggest that you've still got code in here that's specific to stuff and interpreting things that *ideally* are interpreted only in snappl.

Is there an issue about this already?

At some point one of us should go through campari and find all the things that do different things based on what type of PSF you have and so forth, and see if we can make it more general.

ColeFMeldorf: Ah well good news, it seems that this isn't used anywhere and I just forgot to delete it. Problem solved :)).

ColeFMeldorf: For some reason my comment didn't load. This is not used anywhere, so I deleted it!

PR #189, opened 2025-10-07
Removed dev/null and reduced memory

### LINKED ISSUE #180: Unnecessary memory allocation when making blank images.
....that's a lot of memory to allocate to not use.  I need to dig into the code to see how these are used, but the comment you have here suggest that there's a design problem somewhere, either in campari, snappl, or both.  Don't deal with this PR, because the issue was already there, but make a campari Issue that point to this line and says that you're allocating 382MB of memory that is never used, and that there must be a better way.

_Originally posted by @rknop in https://github.com/Roman-Supernova-PIT/campari/pull/177#discussion_r2388282849_

ColeFMeldorf:
Blank images were being generated with large (4088x4088) arrays of zeros that were a waste of memory. 

- [x] To somewhat fix this issue, I've made it such that the blank arrays are created, the cutout performed, and then immediately deleted. This way, the memory usage is reduced by a factor of the number of simulated images. I can't make the arrays smaller or nonexistent because then the cutout will fail. 
- [x] While touching these lines of code, I also removed `/dev/null `s and replaced them with `None`s as I was able to make those changes in `snappl` simultaneously.

This PR is dependent on snappl PR [81](https://github.com/Roman-Supernova-PIT/snappl/pull/81)


wmwv (changes requested):
Update changelog message to be more correct in its description in the updated context.

campari/simulation.py
```diff
@@ -140,6 +140,9 @@ def simulate_images(image_list=None, diaobj=None,
         input_header["BAND"] = image_object.band
         input_header["POINTING"] = image_object.pointing
         input_header["SCA"] = image_object.sca
+        image_object.data = np.zeros((4088, 4088))
```
wmwv: Define 4088 somewhere as a variable and use that consistently.

Possibly something like

nx, ny = 4088, 4088

ColeFMeldorf:
No longer necessary since campari does not simulate its own images.

PR #190, opened 2025-10-08
Moving PSF options to config and removing psfclass from main code

ColeFMeldorf:
In this PR we resolve two issues at once by removing unnecessary specific references to psfclass and by moving PSF arguments to the config file.
To do so:

- [x] Every instance of `psfclass` from the command line is replaced by `transient_psfclass` and/or `galaxy_psfclass` as per Rob's suggestion
- [x] The source_phot_ops in the config has been replaced with transient_photon_ops and galaxy_photon_ops for clarity. 
- [x] Every time psfclass was referenced outside of a call to get_psf_object has been removed. The one exception is in  `build_and_save_lightcurve` where it is used to decide what to name the file. This is to make sure that test output files have the same names as before so tests don't break. If we really hate this, I can go through and change the file names and remove the if statement at line 359 of campari_runner. 
- [x] Same as above but for phot ops.
- [x] There is also still one direct reference to psfclass in line 171 of `simulation.py`, but as of now this is unavoidable, since the snappl PSF objects can not simulate galaxies, I need to switch to galsim to simulate galaxies.

### DIFF
--- a/campari/tests/test_runner.py
+++ b/campari/tests/test_runner.py
@@ -78,6 +78,7 @@ def create_default_test_args(cfg):
     test_args.sim_galaxy_offset_all = config.value("photometry.campari_simulations.sim_galaxy_offset")
     test_args.source_phot_ops = config.value("photometry.campari.source_phot_ops")
     test_args.transient_photon_ops = config.value("photometry.campari.psf.transient_photon_ops")
+    # This will need to go away once the PSF object is split in phot ops and non phot ops
     test_args.galaxy_photon_ops = config.value("photometry.campari.psf.galaxy_photon_ops")
     test_args.mismatch_seds = config.value("photometry.campari_simulations.mismatch_seds")
     test_args.fetch_SED = config.value("photometry.campari.fetch_SED")


campari/tests/test_campari.py
```diff
@@ -583,11 +586,23 @@ def test_construct_transient_scene():
 
     comparison_image = np.load(pathlib.Path(__file__).parent
                                / "testdata/test_psf_source.npy")
+    cfg = Config.get()
+    orig_transient_photops = cfg.value( "photometry.campari.psf.transient_photon_ops" )
```
rknop: Maybe put a comment here saying that this will need to change (this config option will need to go away) when Snappl Issue https://github.com/Roman-Supernova-PIT/snappl/issues/152 is resolved.

campari/tests/test_runner.py
```diff
@@ -77,6 +77,8 @@ def create_default_test_args(cfg):
     test_args.sim_galaxy_scale_all = config.value("photometry.campari_simulations.sim_galaxy_scale")
     test_args.sim_galaxy_offset_all = config.value("photometry.campari_simulations.sim_galaxy_offset")
     test_args.source_phot_ops = config.value("photometry.campari.source_phot_ops")
+    test_args.transient_photon_ops = config.value("photometry.campari.psf.transient_photon_ops")
```
rknop: Same comment as above re: Snappl issue 152.

PR #191, opened 2025-10-13
Updating to the requested lc format!

ColeFMeldorf:
In the last photometry meeting, we discussed updating phrosty and campari to match the lc format requested in the wiki. This was essentially column names and parquet formatting. To do this I:

- [x] Updated column names in lc saving funcs and in saved lc files for tests, and updated filter --> band where it is referenced in tests.
- [x] Added a new test, `test_savelightcurve_pq` which is identical to `test_savelightcurve` other than it writes out a parquet file.
- [x] Similarly, `test_build_and_save_lc_parquet` is the same as `test_build_and_save_lc` but with parquet file type.
- [x] Added a new config file argument, `lc_filetype` which controls the file extension for the output lc. I did this rather than switching to parquet hardcoded because 1.) I'd have to recreate a ton of saved files for tests and 2.) Rob said the definition of what the lc standard is is a moving target so I am staying flexible.
- [x] This argument is passed to `save_lightcurve` which then saves it as the appropriate type.

NOTE: I am not sure as of yet how to define iau_name, provenance_id, or ra_err and dec_err. For now, these all have placeholder values. This PR can be held until we know how to make these if preferred.

### DIFF
--- a/changes/191.campari.rst
+++ b/changes/191.campari.rst
@@ -1,0 +1,1 @@
+Updated lc file output to match the Wiki.


ColeFMeldorf:
Sorry I didn't see this comment! Yes that PR does all of this. Closing!

PR #192, opened 2025-10-14
More Gaussian Tests! (A slight backtrack to where tests pass). 

ColeFMeldorf:
We now have a test for every possible combo of:
Images aligned or unaligned (2 options ) X noise type (4 options, sky, poisson, both or neither) for a total of 8 tests.

For noiseless tests we check that the algorithm recovers truth flux to 1e-7, and we also check that the flux error is a small value. It's not quite zero due to the error_floor that prevents pixels from having zero error.
For Poisson noise only tests we check that errors are within 3 sigma of truth, I check that the number of sub one sigma error is approximately 68%, and I check that the results agrees with aperture photometry.
Finally, for images with sky noise (only sky and both sky and poisson) I check that errors are within 3 sigma of truth, I check that the number of sub one sigma error is approximately 68%, but do not perform the aperture photometry checks, as they are quite different, which is to be expected in the low flux + sky noise case.
ISSUES:
As it stands, the errors measured by campari and aperture phot are a little different in the low flux regime, so I currently do not enforce a check. I could perform a check that just ensures they are within, say, 5 counts of eachother, which would pass, but is this too loose?

To get these tests to work I make the following changes since the last gaussian test PR:

- [x] Update image file paths to match with the more organized photometry test data file structure.
- [x] Regression tests are updated to reflect more accurate sigma_flux values.
- [x] Made the variance of the gaussian weighting a passable argument. This is _not_ used in the new accuracy tests, but it was used in some regression tests, so this is implemented so that those old tests can pass while the new ones can have the gaussian weighting turned off.
- [x] Added a diagnostic image output for the partial overlap test and for some of the gaussian tests.
- [x] Also added command line arguments for the `error_floor `and `cutoff `parameters that go into `get_weights`.
- [x] Updated gitignore to ignore diagnostic images and old out of date regression testdata files.

### DIFF
--- a/changes/192.campari.rst
+++ b/changes/192.campari.rst
@@ -1,0 +1,1 @@
+Adding more gaussian PSF accuracy checks!


ColeFMeldorf:
Superseded by PR 193

PR #193, opened 2025-10-21
Gaussian Tests Part 3: The Reckoning

ColeFMeldorf:
With this PR, campari passes all of the gaussian PSF tests with all of the possible noise sources. I.e. campari pull values look normal with sigma 1 and mu ~ 0 on images that A.) have a host B.) have poisson noise C.) have sky noise and D.) Are offset from each other. I could hypothetically make the host brighter, which I could also try next, but I wanted to hit a checkpoint in my progress. NOTE: This PR looks  massive, but that's because I had to add a lot of testdata to run the tests on. Anything in the testdata subfolder can be safely ignored by my poor reviewers. 

All tests passing except one, which won't pass until we add the ability to draw non point sources with our PSF objects. This test is in test_simulations.py, which is largely being phased out thanks to Rob's imsim, so this is of little concern.

Originally my philosophy was to have a test for every possible combination of noise types and host / no host and rotation / no rotation. However, this was taking a long time. So, while there is a test that runs campari on images with all possible annoyances, I did not, say, do one with every except not sky noise. That seemed like a waste of time when if it can do a harder case it should be able to do an easier case. All in all there are 11 gaussian PSF tests. To get these working I:

- [x] Created a function for the gaussian tests called perform_gaussianity_checks. This function checks that the pulls are gaussian by checking that their mu ~ 0 and sigma ~ 1 and they are not skewed.
- [x] Some tests that were not passing were resolved by increasing the number of images. The small number statistics previously made them fail.
- [x] Tests with a grid now load a prebuilt saved grid model. This cuts the runtime of these tests from ten hours to ten minutes. Of course, if the grid or PSF model is changed, these will also need to be updated.
- [x] Added tests that fit a host without without a transient. This is useful, because there is some noise inherent from the fact that our model is granular and the host is continuous. With a spacing of 5 x 5 grid of points placed over 4 x 4 pixels, the noise solely from the discreteness of the model is well under 20 counts per pixel, subdominant to instrumental / natural sources of noise.
- [x] Many gaussian PSFs with hosts are added, up to and including the hardest test as mentioned in the top of this comment.
- [x] Small lints in `campari/tests/test_runner.py`
- [x] Subsize, the size of the grid which is allowed to be smaller than the image itself, is finally codified as a passable parameter rather than being hardcoded.
- [x] In `campari/campari_runner.py` subsize is added as a param, and I add a +/- 1 on the dates associated with simulated images in order to not confuse the user if a MJD lands exactly on the start date of a transient.
- [x] In `campari/model_building.py` the subsize param is added.
- [x] In `campari/run_one_object.py` I enforce the image sorting order. I was getting all sorts of confusing errors when I now have post transient images. This image sorting enforces the order that campari expects which is no transient then with transient images, and those groups are sorted by MJD.
- [x] Had the code spit out where PSF models are being saved for my sanity.
- [x]  Some useful debug statements added in `campari/utils.py`.
- [x] Config file updated.

LINGERING ISSUES:
1. ) We need to do this with non delta function hosts. I think delta functions will be harder to model than smoother sources, but I also think the delta function galaxy influences the transient less. Unclear if smoothing the galaxy will end up with worse or better fit.
2. ) While residuals are normal and unskewed, some seem to have some structure. I.e. a little bit of an "M" shaped pattern in the lightcurve residuals. Pic included for an example. I know that rolling average seems noisy, but I swear I keep seeing a rolling average that goes up down up like that....
image

3.) How will non linearity affect this?
4.) Can we recreate these results with a roman PSF?

One more note: this PR uses the u/cole/staticsources snappl branch.

### DIFF
--- a/campari/tests/test_campari.py
+++ b/campari/tests/test_campari.py
@@ -430,8 +430,9 @@ def test_make_regular_grid():
                          -44.263774, -44.263868, -44.263796, -44.263724])
     wcs_dict = fits.Header(wcs_dict)
     for wcs in [snappl.wcs.AstropyWCS.from_header(wcs_dict)]:
-        img = FITSImageStdHeaders(header=wcs_dict, path="/dev/null", data=np.zeros((25, 25)))
-        ra_grid, dec_grid = make_regular_grid(img, spacing=3.0, subsize=9)
+        img = FITSImageStdHeaders(header=fits.Header(wcs_dict), path="/dev/null", data=np.zeros((25, 25)))
+        ra_grid, dec_grid = make_regular_grid(img,
+                                              spacing=3.0)
         np.testing.assert_allclose(ra_grid, test_ra, atol=1e-9), \
             "RA vals do not match"
         np.testing.assert_allclose(dec_grid, test_dec, atol=1e-9), \
@@ -452,7 +453,7 @@ def test_make_adaptive_grid():
                                  / "testdata/images.npy")
         SNLogger.debug(f"compare_images shape: {compare_images.shape}")
         image = compare_images[0].reshape(11, 11)
-        img_obj = FITSImageStdHeaders(header=wcs_dict, data=image, path="/dev/null")
+        img_obj = FITSImageStdHeaders(header=fits.Header(wcs_dict), data=image, path="/dev/null")
         ra_grid, dec_grid = make_adaptive_grid(img_obj, percentiles=[99])
         test_ra = [7.67356034, 7.67359491, 7.67362949, 7.67366407, 7.67369864,]
         test_dec = [-44.26425446, -44.26423765, -44.26422084, -44.26420403,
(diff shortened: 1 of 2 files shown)


campari/data_construction.py
```diff
@@ -201,19 +239,14 @@ def find_all_exposures(
...
     image_selection_start, image_selection_end: floats, the first and last MJD of images to be used in the algorithm.
-    explist: astropy.table.Table, the table of exposures that contain the
-    supernova. The columns are:
-        - pointing: the pointing of the exposure
-        - sca: the SCA of the exposure
-        - band: the band of the exposure
-        - date: the MJD of the exposure
-        - detected: whether the exposure contains a detection or not.
+    image_collection: str, the source of the images to be used. If "ou2024", use the Open Universe 2024 images.
+    image_path: str, the path to the images to be used. If given, will use these images
```
rknop: Again, this docstring needs to be a lot longer to explain what this really is.  The current docstring is impenetrable.

PR #195, opened 2025-10-27
Provenances

ColeFMeldorf:
In this PR we implement the specific provenance and database updates that have been rolled out in snappl over the past few weeks. To do so we:

- [x] Update a lot of regression test files to use new col names
- [x] snpit_utils ---> snappl everywhere
- [x] I write a new function, `compare_lightcurve` that compares two lightcurves, for use within tests.
- [x] Tests around lightcurve io have some parameters updated to match new lc formalism
- [x] --image_source is added as a new parameter
- [x] id ---> name in all find_objects calls
- [x] Calls to snpitdb_client added where necessary
- [x] provenance_tag and process arguments added to functions that now need them
- [x] Commented out test_deltafcn_galaxy as it won't work until after Nov meeting
- [x] The subtracted bg flux is now saved during the background subtraction step so that it can go into the final lc.
- [x] Nones are replaced with zeros or 1e30s in find_all_exposures as Nones cause database issues for now.
- [x] campari provenance is determined before saving lightcurve, I started campari at version 0.42.0 to match phrosty, and just for fun.
- [x] params that determine the provenance are now dumped to the meta data of lc
- [x] build_lightcurve is updated to match new snappl Lightcurve inputs
- [x] config paths updated

NOTE: Some tests still use ou2024 as their image source, which I know can't stay. This is due to some oversight on my end where I accidentally made these tests source reliant. They will be fixed but this PR is already huge. Issue is already raised on this matter.

### DIFF
--- a/campari/campari_runner.py
+++ b/campari/campari_runner.py
@@ -166,7 +166,8 @@ class campari_runner:
         # ra=self.ra, dec=self.dec
         #    mjd_discovery_min=self.transient_start, mjd_discovery_max=self.transient_end
 
-        SNLogger.debug(f"Searching for DiaObject with id={self.diaobject_name}, ra={self.ra}, dec={self.dec},"
+        SNLogger.debug(f"Searching for DiaObject with id={self.diaobject_id}, name={self.diaobject_name},"
+                       f" ra={self.ra}, dec={self.dec},"
                        f" collection={self.diaobject_collection}, provenance_tag={self.diaobject_provenance_tag}, "
                        f"process={self.diaobject_process}")
 
@@ -178,9 +179,13 @@ class campari_runner:
                                          mjd_discovery_max=self.transient_end)
 
         if len(diaobjs) == 0:
-            raise ValueError(f"Could not find DiaObject with id={self.diaobject_name}, ra={self.ra}, dec={self.dec}.")
+            raise ValueError(
+                f"Could not find DiaObject with id={self.diaobject_id}, name={self.diaobject_name},"
+                f" ra={self.ra}, dec={self.dec}."
+            )
         if len(diaobjs) > 1:
-            raise ValueError(f"Found multiple DiaObject with id={self.diaobject_name}, ra={self.ra}, dec={self.dec}.")
+            raise ValueError(f"Found multiple DiaObject with id={self.diaobject_id}, name={self.diaobject_name},"
+                             f" ra={self.ra}, dec={self.dec}.")
         diaobj = diaobjs[0]
 
         # Get diaobject position using different methods depending on provenance.
@@ -188,6 +193,9 @@ class campari_runner:
             diaobj.ra = diaobj.ra
             diaobj.dec = diaobj.dec
         else:
+            if self.ra is not None or self.dec is not None:
+                raise ValueError("Cannot provide ra or dec when also providing diaobject_position_provenance_tag."
+                                 "This would lead to provenance confusion.")
             diaobj.ra, diaobj.dec = diaobj.get_position(provenance_tag=self.diaobject_position_provenance_tag,
                                                         process=self.diaobject_process, dbclient=self.dbclient)
 
@@ -246,13 +254,14 @@ class campari_runner:
             self.pointing_list = None
 
     def create_sim_param_grid(self):
-        """Create a grid of simulation parameters to run the pipeline on."""
-        params = [self.bg_gal_flux_all, self.sim_galaxy_scale_all, self.sim_galaxy_offset_all]
-        nd_grid = np.meshgrid(*params)
-        self.param_grid = np.array(nd_grid, dtype=float).reshape(len(params), -1)
-        SNLogger.debug("Created a grid of simulation parameters with a total of"
-                       f" {self.param_grid.shape[1]} combinations.")
-        self.diaobject_name = self.diaobject_name * self.param_grid.shape[1]  # Repeat the SNID for each combination of parameters
+        raise NotImplementedError("Simulation parameter grid creation is broken. Will be made external later.")
+        # """Create a grid of simulation parameters to run the pipeline on."""
+        # params = [self.bg_gal_flux_all, self.sim_galaxy_scale_all, self.sim_galaxy_offset_all]
+        # nd_grid = np.meshgrid(*params)
+        # self.param_grid = np.array(nd_grid, dtype=float).reshape(len(params), -1)
+        # SNLogger.debug("Created a grid of simulation parameters with a total of"
+        #                f" {self.param_grid.shape[1]} combinations.")
+        # self.diaobject_name = self.diaobject_name * self.param_grid.shape[1]  # Repeat the SNID for each combination of parameters
 
     def get_exposures(self, diaobj):
         """Call the find_all_exposures function to get the exposures for the given RA, Dec, and time frame."""
--- a/campari/data_construction.py
+++ b/campari/data_construction.py
@@ -230,6 +230,11 @@ def find_all_exposures(
     temp_transient_start = 0 if transient_start is None else transient_start
     temp_transient_end = 1e30 if transient_end is None else transient_end
 
+    SNLogger.debug(f"Using image collection: {image_collection}")
+    SNLogger.debug(f"Using provenance tag: {provenance_tag}")
+    SNLogger.debug(f"Using process: {process}")
+    SNLogger.debug(f"db_client: {dbclient}")
+
     img_collection = ImageCollection().get_collection(collection=image_collection, provenance_tag=provenance_tag,
                                                       process=process, dbclient=dbclient)
 
--- a/campari/io.py
+++ b/campari/io.py
@@ -71,7 +71,7 @@ def build_lightcurve(diaobj, lc_model, obj_pos_prov=None):
     cam_prov = Provenance(
         process="campari",
         major=0,
-        minor=42,
+        minor=42, # THIS CAN'T BE HARDCODED FOREVER XXX TODO
         params=cfg,
         keepkeys=["photometry.campari"],
         omitkeys=None,
@@ -102,6 +102,7 @@ def build_lightcurve(diaobj, lc_model, obj_pos_prov=None):
     for i, img in enumerate(image_list):
         if img.mjd > diaobj.mjd_start and img.mjd < diaobj.mjd_end:
             data_dict["mjd"].append(img.mjd)
+            SNLogger.debug(f"Adding image with MJD {img.mjd} to lightcurve")
             data_dict["pointing"].append(img.pointing)
             data_dict["sca"].append(img.sca)
             x, y = img.get_wcs().world_to_pixel(diaobj.ra, diaobj.dec)
@@ -125,7 +126,13 @@ def build_lightcurve(diaobj, lc_model, obj_pos_prov=None):
     # Note that this is only allowing for one band, not multiple bands. I don't think campari will ever
     # do multi-band fitting so this is probably fine.
     meta_dict[f"local_surface_brightness_{band}"] = lc_model.LSB
-    return Lightcurve(data=data_dict, meta=meta_dict)
+
+    lc = Lightcurve(data=data_dict, meta=meta_dict)
+    # Some extra info needed to save
+    lc.image_list = image_list
+    lc.diaobj = diaobj
+
+    return lc
 
 
 def build_lightcurve_sim(supernova, flux, sigma_flux):
@@ -148,7 +155,10 @@ def build_lightcurve_sim(supernova, flux, sigma_flux):
     return QTable(data=data_dict, meta=meta_dict, units=units)
 
 
-def save_lightcurve(lc=None, identifier=None, psftype=None, output_path=None, overwrite=True, save_to_database=False):
+def save_lightcurve(lc=None, identifier=None, psftype=None, output_path=None,
+                    overwrite=True, save_to_database=False, dbclient=None,
+                    new_provenance=False, diaobj_pos=None,
+                    ltcvprocess=None):
     """This function parses settings in the SMP algorithm and saves the
     lightcurve to an ecsv file with an appropriate name.
     Input:
@@ -183,9 +193,38 @@ def save_lightcurve(lc=None, identifier=None, psftype=None, output_path=None, ov
 
     filepath = f"{identifier}_{band}_{psftype}_lc.ecsv" if not save_to_database else None
 
-    lc.write(
-        base_dir=output_path, filepath=filepath, filetype="ecsv", overwrite=overwrite
-    )
+    if save_to_database:
+
+        # process = "campari"
+        # major = 0
+        # minor = 42  # THIS CAN'T BE HARDCODED FOREVER XXX TODO
+        # params = Config.get()
+        # keepkeys = ["photometry.campari"]
+        # omitkeys = None
+        # imgprov = Provenance.get_by_id(lc.image_list[0].provenance_id, dbclient=dbclient)
+        # objprov = Provenance.get_by_id(lc.diaobj.provenance_id, dbclient=dbclient)
+        # if diaobj_pos is not None:
+        #     objposprov = Provenance.get_by_id(diaobj_pos, dbclient=dbclient)
+        #     upstreams = [imgprov, objprov, objposprov]
+        # else:
+        #     objposprov = None
+        #     upstreams = [imgprov, objprov]
+        SNLogger.debug("Creating new provenance for lightcurve")
+
+        # ltcvprov = Provenance(process=process, major=major, minor=minor, params=params,
+        #                       upstreams=upstreams, keepkeys=keepkeys,
+        #                       omitkeys=omitkeys)
+        SNLogger.debug("Using provenance to save lightcurve to database")
+        SNLogger.debug(lc.meta.get("provenance_id"))
+        ltcvprov = Provenance.get_by_id(lc.meta.get("provenance_id"), dbclient=dbclient)
+        ltcv_provenance_tag = "campari_test_provenance_lightcurve"
+        if new_provenance:
+            ltcvprov.save_to_db(tag=ltcv_provenance_tag)
+        lc.save_to_db(dbclient=dbclient)
+    else:
+        lc.write(
+            base_dir=output_path, filepath=filepath, filetype="ecsv", overwrite=overwrite
+        )
     # Return the lc so we can have the snappl generated filepath
     return lc
 
--- a/campari/tests/test_database.py
+++ b/campari/tests/test_database.py
@@ -13,6 +13,8 @@ from erfa import ErfaWarning
 
 # SNPIT
 from campari.data_construction import find_all_exposures
+from campari.io import build_lightcurve, save_lightcurve
+from campari.utils import campari_lightcurve_model
 
 from snappl.dbclient import SNPITDBClient
 from snappl.diaobject import DiaObject
@@ -74,14 +76,17 @@ def test_get_image_collection_missing_provenance():
 
 
 def test_find_exposures():
-    provenance_tag = "ou2024"
-    process = "load_ou2024_image"
+
     dbclient = SNPITDBClient()
 
     diaobj_provenance_tag = "ou2024"
     diaobj_process = "load_ou2024_diaobject"
     diaobj = DiaObject.find_objects(name=20172782, collection="snpitdb",
                                     provenance_tag=diaobj_provenance_tag, process=diaobj_process)[0]
+
+    provenance_tag = "ou2024"
+    process = "load_ou2024_image"
+
     diaobj.mjd_start = 62654.0
     diaobj.mjd_end = 62958.0
     image_list, _ = find_all_exposures(diaobj=diaobj, band="Y106",
@@ -93,3 +98,4 @@ def test_find_exposures():
     pointings = [i.pointing for i in image_list]
     regression_pointings = np.load("/campari/campari/tests/testdata/test_find_exposures_pointings.npy")
     np.testing.assert_array_equal(pointings, regression_pointings)
+
--- a/campari/tests/test_runner.py
+++ b/campari/tests/test_runner.py
@@ -302,23 +302,24 @@ def test_build_and_save_lc(cfg):
     compare_lightcurves(current, comparison)
 
 
-def test_sim_param_grid(cfg):
-    test_args = create_default_test_args(cfg)
-    test_args.use_real_images = False
-    test_args.diaobject_collection = "ou24"
-    test_args.diaobject_name = 20172782
-    runner = campari_runner(**vars(test_args))
-    runner.decide_run_mode()
-    runner.bg_gal_flux_all = [1.0, 2.0]
-    runner.sim_galaxy_scale_all = [1.0, 2.0, 3.0]
-    runner.sim_galaxy_offset_all = 0.0
-    # Create the simulation parameter grid
-    runner.create_sim_param_grid()
+# sim param grid broken for now
+# def test_sim_param_grid(cfg):
+#     test_args = create_default_test_args(cfg)
+#     test_args.use_real_images = False
+#     test_args.diaobject_collection = "ou24"
+#     test_args.diaobject_name = 20172782
+#     runner = campari_runner(**vars(test_args))
+#     runner.decide_run_mode()
+#     runner.bg_gal_flux_all = [1.0, 2.0]
+#     runner.sim_galaxy_scale_all = [1.0, 2.0, 3.0]
+#     runner.sim_galaxy_offset_all = 0.0
+#     # Create the simulation parameter grid
+#     runner.create_sim_param_grid()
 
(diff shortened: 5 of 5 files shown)


campari/data_construction.py
```diff
@@ -213,34 +215,60 @@ def find_all_exposures(
...
+    temp_image_selection_end = 1e30 if image_selection_end is None else image_selection_end
+    temp_transient_start = 0 if transient_start is None else transient_start
+    temp_transient_end = 1e30 if transient_end is None else transient_end
+
+    img_collection = ImageCollection().get_collection(collection=image_source, provenance_tag=provenance_tag,
+                                                      process=process, dbclient=dbclient)
+    if image_source == "ou2024":
+        img_collection_prov = None
+    else:
+        img_collection_prov = img_collection.provenance
```
rknop: Also, as a general rule -- if you find yourself interpreting the value of `image_source` (by which I assume you mean the collection?), you're doing it wrong.

Things like this need to be opaque to campari, and just passed on to snappl.  If snappl doesn't fully support what you need, we need to fix it there.  But, the more you code in things in campari that look at these values, the more trouble we're going to have in the future.

If you MUST, just to get things done, you need to put in commens that this is a bad and needs to be fixed, and it needs to be referenced in an issue (making a new issue if there isn't one already).

campari/io.py
```diff
@@ -58,42 +62,79 @@ def build_lightcurve(diaobj, lc_model):
...
+    provs = [diaobj_prov, lc_model.image_collection_prov]
+    upstream_list = [p for p in provs if p is not None]
+
+    cfg = Config.get()
+    cam_prov = Provenance(
+        process="campari",
+        major=0,
+        minor=42,
+        params=cfg,  # keepkeys=["photometry.campari"],
+        omitkeys=["photometry.campari.galsim", "photometry.campari.simulations"],
```
rknop: This isn't what you want, because as is right now, you will include all of the `system` tree, which is explicitly what you don't want

What you want is `keepkeys=['photometry.campari']`.  If there are then things unerneath `photometry.campari` that aren't supposed to be part of the config, they should be moved to somewhere else.  (What's in `phtoometry.campari.galsim` and `photometry.camapri.simulations`?  If you are building lightcurves from simulations and you are saving them, you absolutely want those as part of the params.)

campari/RomanASP.py
```diff
@@ -135,11 +135,11 @@ def main():
                              "around the given RA and Dec. If not given, "
                              "will return the closest.")
 
-    parser.add_argument("--object_collection", type=str, default="ou2024", required=False,
+    parser.add_argument("--object_collection", type=str, default="snpitdb", required=False,
                         help="Which collection of objects to use for lookup. "
                              "Default is 'ou24', the Open Universe 2024 catalog. 'manual'"
                              "will use the input ra and dec given by the user, and not perform any lookup.")
-    parser.add_argument("--image_source", type=str, default="ou2024", required=False,
+    parser.add_argument("--image_source", type=str, default="snpitdb", required=False,
```
rknop: As I noted above, you will need to add six more arguments here:

* Image collection  provenance tag and process
* Object provenance tag and process
* Object position provenance tag and process

All of these can default to `None`.  (You will get errors from snappl if you try to get image collections or objects without a provenance when using a database image collection; that's fine, let snappl throw the error, you shouldn't verify this yourself.)

You will want to look inside the object position provenance tag; if that is none, use the RA and Dec from the object.  If not, then get the object position using the method in `DiaObject` that does that.

examples/perlmutter/interactive_podman.sh
```diff
@@ -4,6 +4,7 @@ podman-hpc run --gpu \
     --mount type=bind,source=$PWD,target=/home \
     --mount type=bind,source=$PWD/campari,target=/campari \
     --mount type=bind,source=$PWD/campari_out_dir,target=/campari_out_dir \
+    --mount type=bind,source=$PWD/campari_out_dir,target=/lightcurves \
```
rknop: ... did this actually work?  I don't see a bind-mount for `/secrets` . (I may be blind.)

examples/perlmutter/campari_config_test.yaml
```diff
@@ -53,16 +53,23 @@ photometry:
...
   snappl:
-    simdex_server: https://roman-desc-simdex.rknop.net
-
-ou24:
-  config_file: /sn_info_dir/tds.yaml
-  sn_truth_dir: /snana_pq_dir
-  sims_sed_library: /home/rubin_sim_data/sims_sed_library
-  images: /sims_dir/RomanTDS/images/simple_model
-  tds_base: /photometry_test_data/ou2024
+    simdex_server: https://c3-sn.lbl.gov/roman_snpit_ou2024
```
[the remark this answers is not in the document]
ColeFMeldorf: Done!

ColeFMeldorf:
Note to self: object position prov still needs to be added to upstreams

ColeFMeldorf:
Note to self: make sure I can write to parquet files

campari/io.py
```diff
@@ -132,14 +173,28 @@ def save_lightcurve(lc=None, identifier=None, psftype=None, output_path=None, ov
...
-    SNLogger.info(f"Saving lightcurve to {lc_file}")
-    lc.write(lc_file, format="ascii.ecsv", overwrite=overwrite)
+    band_map = {
+            "r": "R062",
+            "z": "Z087",
+            "y": "Y106",
+            "j": "J129",
+            "h": "H158",
+            "f": "F184",
+            "w": "W146",
```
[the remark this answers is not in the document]
ColeFMeldorf: I realized I made a mistake and had a [0] after band, just giving a single letter. I mistakenly thought that meant Lightcurve was abbreviating bands. This has been removed and the dictionary removed as well.

campari/campari_runner.py
```diff
@@ -169,7 +195,7 @@ def __call__(self):
             SNLogger.debug(f"Object info for SN {ID} in collection {self.object_collection}: ra={diaobj.ra},"
                            f" dec={diaobj.dec}, transient_start={diaobj.mjd_start}, transient_end={diaobj.mjd_end}")
             image_list = self.get_exposures(diaobj)
-            sedlist = self.get_sedlist(diaobj.id, image_list)
+            sedlist = self.get_sedlist(diaobj.name, image_list)
```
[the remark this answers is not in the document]
rknop: Ultimately we're going to have to think about getting object SEDs from snappl, I think.  We need to sit down as photometry in general and talk about how we're going to handle SEDs, because it's nontrivial when you don't have truth files.

campari/campari_runner.py
```diff
@@ -362,27 +410,28 @@ def call_run_one_object(self, diaobj, image_list, sedlist, param_grid_row):
...
+
+        lc_model.image_collection_prov = self.img_coll_prov if self.use_real_images else None
         if self.psfclass == "ou24PSF" or self.psfclass == "ou24PSF_slow":
             psftype = "romanpsf"
         else:
             psftype = self.psfclass.lower()
 
         if self.use_real_images:
-            identifier = str(diaobj.id)
+            identifier = str(diaobj.name)
```
rknop: What do you use this identifier for later?

As I've noted a few times `DiaObject.name` is not a reliable property.  There may be duplications, some objects will have no name.

The safest way to uniquely identify a `DiaObject` is with its id, at least if it came from the database.  If it didn't, then things are complicated.  We should think about this.

You might do `identifier = str(diaobj.id if diaobj.id is not None else diaobj.name)`... but that depends on what you're doing with `identifier` later.

campari/io.py
```diff
@@ -57,43 +60,82 @@ def build_lightcurve(diaobj, lc_model):
...
+    meta_dict["provenance_id"] = cam_prov.id
+    meta_dict["diaobject_id"] = diaobj.id
+    meta_dict["iau_name"] = diaobj.name  # I am not sure this is what IAUname is but it's a placeholder for now.
+    meta_dict["ra_err"] = 0.0
+    meta_dict["dec_err"] = 0.0  # This is a placeholder for now
+    meta_dict["ra_dec_covar"] = 0.0  # This is a placeholder for now
+    # Note that this is only allowing for one band, not multiple bands. I don't think campari will ever
+    # do multi-band fitting so this is probably fine.
+    meta_dict[f"local_surface_brightness_{band}"] = lc_model.LSB
+    data_dict["NEA"] = [0.0] * len(data_dict["pix_x"])  # snappl will calculate this
```
rknop: Oh... you've got that already. Ignore previous comment.

campari/RomanASP.py
```diff
@@ -215,6 +215,30 @@ def main():
...
+    parser.add_argument("--get_collection_prov_tag", "--get_collection_provenance_tag", type=str,
+                        help="A string tag to identify the provenance of "
+                        "the image collection step. Default None.", default=None)
+
+    parser.add_argument("--obj_pos_prov_tag", "--object_position_provenance_tag", type=str,
+                        help="A string tag to identify the provenance of the program that determines the "
+                        "object position. Default None.", default=None)
+
+    parser.add_argument("--obj_pos_process", "--object_position_process", type=str,
+                        help="A string to identify the process of the program that determines the object position."
```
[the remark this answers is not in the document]
ColeFMeldorf: Check! All included now.

campari/campari_runner.py
```diff
@@ -267,16 +296,25 @@ def create_sim_param_grid(self):
...
+                                                                maxbg=self.max_no_transient_images,
+                                                                maxdet=self.max_transient_images,
+                                                                band=self.band,
+                                                                image_selection_start=self.image_selection_start,
+                                                                image_selection_end=self.image_selection_end,
+                                                                image_source=self.image_source,
+                                                                pointing_list=self.pointing_list,
+                                                                dbclient=self.dbclient,
+                                                                collection_provenance_tag=self.
+                                                                get_collection_provenance_tag,
```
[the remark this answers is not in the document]
ColeFMeldorf: check!

campari/campari_runner.py
ColeFMeldorf: I was using identifier to create the filename for the output lightcurve later. The filename is essentially identifier + band +etc. ecsv, if the id is a super long hash, I am not sure if it makes for good lc file names, unless you think otherwise?

campari/io.py
```diff
@@ -132,14 +174,15 @@ def save_lightcurve(lc=None, identifier=None, psftype=None, output_path=None, ov
     The file name is:
     <output_path>/identifier_band_psftype_lc.ecsv
     """
-    band = lc["filter"][0]
+    band = lc.meta["band"]
+    SNLogger.debug(f"saving lightcurve for id={identifier}, band={band}, psftype={psftype}")
     output_path = Config.get().value("photometry.campari.paths.output_dir") if output_path is None else output_path
```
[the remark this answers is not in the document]
ColeFMeldorf: Done and done!

campari/io.py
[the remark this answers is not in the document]
rknop: For tests, that's fine, though longer term we're going to want to make it somewhere else.  We'll do that when we set up your tests to have a self-contained environment.  (Did you make an issue for that?)  This is post-November stuff.

campari/io.py
```diff
@@ -132,14 +174,15 @@ def save_lightcurve(lc=None, identifier=None, psftype=None, output_path=None, ov
...
     output_path = pathlib.Path(output_path)
     output_path.mkdir(exist_ok=True, parents=True)
 
-    lc_file = output_path / f"{identifier}_{band}_{psftype}_lc.ecsv"
-    SNLogger.info(f"Saving lightcurve to {lc_file}")
-    lc.write(lc_file, format="ascii.ecsv", overwrite=overwrite)
+    lc.write(
+        base_dir=output_path, filepath=f"{identifier}_{band}_{psftype}_lc.ecsv", filetype="ecsv", overwrite=overwrite
```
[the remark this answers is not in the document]
ColeFMeldorf: As per our slack conversation, this now only has a filepath if not saving to the database!

campari/io.py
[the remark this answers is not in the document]
ColeFMeldorf: I have now!

campari/campari_runner.py
```diff
@@ -362,34 +436,36 @@ def call_run_one_object(self, diaobj, image_list, sedlist, param_grid_row):
...
         else:
             sim_galaxy_scale, bg_gal_flux, sim_galaxy_offset = param_grid_row
             if self.run_name is None:
                 identifier = "simulated_" + str(sim_galaxy_scale) + "_" + \
                     str(np.round(np.log10(bg_gal_flux), 2)) + "_" + str(sim_galaxy_offset) + "_" \
                     + self.grid_type
             else:
-                identifier = self.run_name + "_" + str(diaobj.id)
+                identifier = self.run_name + "_" + str(diaobj.id if diaobj.id is not None else diaobj.name)
             if lc_model.flux is not None:
```
rknop: It's not obvious to me that `lc_model.flux` is a proxy for "we are building a simulated lightcurve".  Is that in fact what it is?

campari/io.py
```diff
@@ -57,43 +60,72 @@ def build_lightcurve(diaobj, lc_model):
...
-    return QTable(data=data_dict, meta=meta_dict, units=units)
+            data_dict["sky_background"].append(lc_model.sky_background[i])
+            data_dict["sky_rms"].append(0.0)  # placeholder for now XXX TODO
+            data_dict["NEA"].append(0.0)  # placeholder for now XXX TODO
+
+    meta_dict["band"] = band  # I don't ever expect campari to do multi-band fitting so just store the one band.
+    meta_dict["diaobject_position_id"] = None  # placeholder for now XXX TODO
+    meta_dict["provenance_id"] = cam_prov.id
+    meta_dict["diaobject_id"] = diaobj.id
+    meta_dict["iau_name"] = diaobj.name  # I am not sure this is what IAUname is but it's a placeholder for now.
```
rknop: `diaobj.iauname` is a thing, just use that.

campari/io.py
```diff
@@ -116,30 +148,35 @@ def build_lightcurve_sim(supernova, flux, sigma_flux):
...
-    output_path = Config.get().value("photometry.campari.paths.output_dir") if output_path is None else output_path
-    output_path = pathlib.Path(output_path)
-    output_path.mkdir(exist_ok=True, parents=True)
-
-    lc_file = output_path / f"{identifier}_{band}_{psftype}_lc.ecsv"
-    SNLogger.info(f"Saving lightcurve to {lc_file}")
-    lc.write(lc_file, format="ascii.ecsv", overwrite=overwrite)
+    band = lc.meta["band"]
+    SNLogger.debug(f"saving lightcurve for id={identifier}, band={band}, psftype={psftype}")
+    base_output_path = Config.get().value("system.paths.lightcurves") if output_path is None else output_path
```
rknop: As above : you should only use `system.paths.lightcurves` if `save_to_database` is True.

(In fact, also, if save_to_database is true and `output_path` is not `None`, then you should raise an exception.)

ColeFMeldorf: done!

campari/campari_runner.py
```diff
@@ -362,34 +436,36 @@ def call_run_one_object(self, diaobj, image_list, sedlist, param_grid_row):
...
                 lc = build_lightcurve_sim(lc_model.sim_lc, lc_model.flux, lc_model.sigma_flux)
                 lc["filter"] = self.band
 
         if lc_model.flux is not None:
-            output_dir = pathlib.Path(self.cfg.value("photometry.campari.paths.output_dir"))
-            save_lightcurve(lc=lc, identifier=identifier, psftype=psftype, output_path=output_dir)
+            output_dir = pathlib.Path(self.cfg.value("system.paths.lightcurves"))
```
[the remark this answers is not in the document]
rknop: Don't just switch; do the check on `save_to_db`.  Never write to `system.paths.lightcurves` when not saving to db, always write there (and let snappl choose the file path) when saving to db

campari/campari_runner.py
rknop: Put in a comment to that effect.

campari/campari_runner.py
rknop: Delete the entire contents of this function down to `if self.img_list is not None:`

It's not needed, if you also get rid of the other things I mentioned in the other suggested change above.  That will save you from having to do all kinds of stuff here to think about `name` (renamed `SNID`) and `diaobject_id`... none of it's needed, because when you do `DiaObject.find_objects`, if the user hasn't given enough information, `snappl` will raise any necesary errors.  (E.g. if the user asks for image collection `manual` but doesn't give an ra and a dec, snappl will already yell; so, you don't have to check for that.)

campari/data_construction.py
```diff
@@ -213,34 +216,54 @@ def find_all_exposures(
         - detected: whether the exposure contains a detection or not.
     """
     SNLogger.debug(f"Finding all exposures for diaobj {diaobj.mjd_start, diaobj.mjd_end, diaobj.ra, diaobj.dec}")
+    SNLogger.debug(f"Using image source: {image_collection}")
```
rknop:
```suggestion
    SNLogger.debug(f"Using image collection: {image_collection}")
```

examples/perlmutter/campari_config_test.yaml
```diff
@@ -17,8 +17,8 @@ photometry:
     paths:
```
rknop: Get rid of this whole section.  All absolute paths need to be under `system` somewhere.

campari/data_construction.py
```diff
@@ -213,34 +216,54 @@ def find_all_exposures(
...
+    temp_transient_start = 0 if transient_start is None else transient_start
+    temp_transient_end = 1e30 if transient_end is None else transient_end
 
-    if (image_selection_start is None or transient_start > image_selection_start) and transient_start is not None:
+    img_collection = ImageCollection().get_collection(collection=image_collection, provenance_tag=provenance_tag,
+                                                      process=process, dbclient=dbclient)
+
+    img_collection_prov = getattr(img_collection, "provenance", None)
+    if img_collection_prov is None:
+        SNLogger.warning("Image collection has no provenance information. This is only a problem if using snpitdb.")
```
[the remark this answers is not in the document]
rknop: ...actually, you return `img_collection_prov`, so don't delete that line.  But you can delete the check.

campari/campari_runner.py
```diff
@@ -138,13 +168,34 @@ def __call__(self):
...
+
+            SNLogger.debug(f"Searching for DiaObject with id={ID}, ra={self.ra}, dec={self.dec},"
+                           f" collection={self.diaobject_collection}, provenance_tag={self.diaobject_provenance_tag}, "
+                           f"process={self.diaobject_process}")
+
+            diaobjs = DiaObject.find_objects(collection=self.diaobject_collection, dbclient=self.dbclient,
+                                             provenance_tag=self.diaobject_provenance_tag,
+                                             process=self.diaobject_process, name=ID,
+                                             ra=self.ra, dec=self.dec, mjd_discovery_min=self.transient_start,
+                                             mjd_discovery_max=self.transient_end)
```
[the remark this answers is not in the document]
ColeFMeldorf: I take that this means I should get rid of --healpix and --nside as well?

campari/campari_runner.py
```diff
@@ -166,7 +217,7 @@ def __call__(self):
                                      f"nominal transient_end {diaobj.mjd_end}")
                 diaobj.mjd_end = self.transient_end
 
-            SNLogger.debug(f"Object info for SN {ID} in collection {self.object_collection}: ra={diaobj.ra},"
+            SNLogger.debug(f"Object info for SN {ID} in collection {self.diaobject_collection}: ra={diaobj.ra},"
```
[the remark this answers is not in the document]
ColeFMeldorf: check!

campari/campari_runner.py
ColeFMeldorf: Check!

campari/data_construction.py
ColeFMeldorf: check

campari/campari_runner.py
```diff
@@ -135,113 +160,83 @@ def __call__(self):
...
-
-            lightcurve_model = self.call_run_one_object(diaobj, image_list, sedlist, param_grid_row)
-            self.build_and_save_lightcurve(diaobj, lightcurve_model, param_grid_row)
+        banner(f"Running SN {self.diaobject_name}")
+
+        # These will need to be re included once Issue #93 is resolved.
+        # ra=self.ra, dec=self.dec
+        #    mjd_discovery_min=self.transient_start, mjd_discovery_max=self.transient_end
+
+        SNLogger.debug(f"Searching for DiaObject with id={self.diaobject_name}, ra={self.ra}, dec={self.dec},"
```
rknop: `id={self.diaobject_id}, name={self.diaobject_name}`

Really don't want to mix up id and name.

campari/campari_runner.py
```diff
@@ -135,113 +160,83 @@ def __call__(self):
...
+
+        diaobjs = DiaObject.find_objects(collection=self.diaobject_collection, dbclient=self.dbclient,
+                                         provenance_tag=self.diaobject_provenance_tag,
+                                         process=self.diaobject_process, name=self.diaobject_name,
+                                         diaobject_id=self.diaobject_id,
+                                         ra=self.ra, dec=self.dec, mjd_discovery_min=self.transient_start,
+                                         mjd_discovery_max=self.transient_end)
+
+        if len(diaobjs) == 0:
+            raise ValueError(f"Could not find DiaObject with id={self.diaobject_name}, ra={self.ra}, dec={self.dec}.")
```
rknop: `id={self.diaobject_id}, name={self.diaobject_name}`

Really don't want to mix up id and name.

campari/campari_runner.py
```diff
@@ -135,113 +160,83 @@ def __call__(self):
...
+        if self.ra is not None:
+            if np.fabs(self.ra - diaobj.ra) > 1. / 3600. / np.cos(diaobj.dec * np.pi / 180.):
+                SNLogger.warning(f"Given RA {self.ra} is far from DiaObject nominal RA {diaobj.ra}")
+            diaobj.ra = self.ra
+        if self.dec is not None:
+            if np.fabs(self.dec - diaobj.dec) > 1. / 3600.:
+                SNLogger.warning(f"Given Dec {self.dec} is far from DiaObject nominal Dec {diaobj.dec}")
+            diaobj.dec = self.dec
+
+        if (self.transient_start is not None):
```
rknop: Make in issue ...somewhere... that we need to think about how we're going to handle the whole notion of transient start and end.

Everything you've done so far has been easy because you have truth tables.

It becomes ill-defined otherwise.  If it's always passed on the command line, then that becomes a little scary.  We need to have a better procedure for deciding these in bulk.  This isn't just a campari issue, this is a broarder Photometry issue.

rknop (changes requested):
Getting closer.  There are some log outputs that need to be fixed, plus I *think* there's still a case of something in there that assumes you're looping over a list of objects, even though that loop is now gone.

campari/campari_runner.py
ColeFMeldorf: check

campari/campari_runner.py
```diff
@@ -135,113 +160,83 @@ def __call__(self):
...
+                                         provenance_tag=self.diaobject_provenance_tag,
+                                         process=self.diaobject_process, name=self.diaobject_name,
+                                         diaobject_id=self.diaobject_id,
+                                         ra=self.ra, dec=self.dec, mjd_discovery_min=self.transient_start,
+                                         mjd_discovery_max=self.transient_end)
+
+        if len(diaobjs) == 0:
+            raise ValueError(f"Could not find DiaObject with id={self.diaobject_name}, ra={self.ra}, dec={self.dec}.")
+        if len(diaobjs) > 1:
+            raise ValueError(f"Found multiple DiaObject with id={self.diaobject_name}, ra={self.ra}, dec={self.dec}.")
```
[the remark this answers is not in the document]
ColeFMeldorf: :(( Yeah I really missed all these, good catch

campari/campari_runner.py
```diff
@@ -250,33 +245,36 @@ def decide_run_mode(self):
...
-
     def create_sim_param_grid(self):
         """Create a grid of simulation parameters to run the pipeline on."""
         params = [self.bg_gal_flux_all, self.sim_galaxy_scale_all, self.sim_galaxy_offset_all]
         nd_grid = np.meshgrid(*params)
         self.param_grid = np.array(nd_grid, dtype=float).reshape(len(params), -1)
         SNLogger.debug("Created a grid of simulation parameters with a total of"
                        f" {self.param_grid.shape[1]} combinations.")
-        self.SNID = self.SNID * self.param_grid.shape[1]  # Repeat the SNID for each combination of parameters
+        self.diaobject_name = self.diaobject_name * self.param_grid.shape[1]  # Repeat the SNID for each combination of parameters
```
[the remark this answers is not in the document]
ColeFMeldorf: Yes my plan is to make it external! I think that's the cleanest way to do it. It also may look a little different depending on how advanced we make your image simulator. In the spirit of that, I think for now I'll comment out this code and reintroduce it later. If you prefer the commented code gone just let me know.

PR #203, opened 2025-10-31
All asserts outside of tests removed

ColeFMeldorf:
Rob said to remove outside of tests and I agree. Thankfully, there was only one.


rknop (approved):
...there really weren't many left.

PR #206, opened 2025-11-06
Small fixes to get a November DB Run through.

ColeFMeldorf:
To get that to work I:

- [x] Moved the config around so that only the appropriate params enter into the provenance determination
- [x] Created a new directory, nov_run, for the november plumbing tests.
- [x] Updated file paths so use "system" and not "campari".
- [x] Added some catches to ensure that Nonetype provenances actually are nonetypes and not the string "None"
- [x] Added the testrun param which allows me to run the same lightcurve multiple times by incrementing this number, which changes the provenance.
- [x] Added a nifty fixture to pytest called overwrite_meta. If I pass this, the tests only check that lightcurve data matches the file. If it does, it overwrites the metadata of the file and reruns itself with overwrite_meta = False. This prevents me from spending 10-20 minutes each time manually updating the metadata of lcs when I change one small thing. Super nice!

### DIFF
--- a/campari/io.py
+++ b/campari/io.py
@@ -84,10 +84,6 @@ def build_lightcurve(diaobj, lc_model, obj_pos_prov=None, dbclient=None):
     else:
         SNLogger.warning("No diaobject position provenance ID provided; skipping.")
 
-    if len(upstreams) == 0:
-        upstreams = None
-        SNLogger.warning("No upstream provenances found; setting upstreams to None.")
-
     cfg = Config.get()
     SNLogger.debug("Attempting to build provenance for lightcurve")
     cam_prov = Provenance(
@@ -100,14 +96,10 @@ def build_lightcurve(diaobj, lc_model, obj_pos_prov=None, dbclient=None):
         upstreams=upstreams,
     )
 
-    SNLogger.debug("Lightcurve provenance object line 98:")
-    SNLogger.debug(cam_prov.params)
 
     meta_dict = cam_prov.params["photometry"]["campari"].copy()
     meta_dict.update({"ID": diaobj.name, "ra": diaobj.ra, "dec": diaobj.dec})
 
-    SNLogger.debug("Lightcurve provenance object line 104:")
-    SNLogger.debug(cam_prov.params)
 
     data_dict = {
         "mjd": [],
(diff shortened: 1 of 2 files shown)


campari/io.py
```diff
@@ -209,18 +242,32 @@ def save_lightcurve(lc=None, identifier=None, psftype=None, output_path=None,
...
-        SNLogger.debug("Using provenance to save lightcurve to database")
         SNLogger.debug(lc.meta.get("provenance_id"))
-        ltcvprov = Provenance.get_by_id(lc.meta.get("provenance_id"), dbclient=dbclient)
-        ltcv_provenance_tag = "campari_test_provenance_lightcurve"
+        #ltcvprov = Provenance.get_by_id(lc.meta.get("provenance_id"), dbclient=dbclient)
+        ltcvprov = lc.provenance_object
+        SNLogger.debug("Lightcurve provenance object:")
+        SNLogger.debug(ltcvprov.id)
+
+        ltcv_provenance_tag = "nov2025campari_firstrun_3"
```
rknop: This is something that needs to be passed in on the command line.

(And, indeed, it looks like you have a `--ltcv-provenance_tag` argument.  You should use that.)

If you use that argument, then I think you can get rid of the `testrun` stuff, because you can just give it a different provenance on the command line.  (At least, the `testrun` that's passed to this function.  You may still need it in params to make different provenances... BUT REALLY, as we know, we want to figure out how to get you a self-contained dev environment.)

campari/tests/conftest.py
```diff
@@ -9,3 +10,18 @@
 @pytest.fixture(scope='module')
 def cfg():
     return Config.get()
+
+# Source - https://stackoverflow.com/a
+# Posted by clay, modified by community. See post 'Timeline' for change history
+# Retrieved 2025-11-06, License - CC BY-SA 4.0
+
+def pytest_addoption(parser):
+    parser.addoption("--overwrite_meta", action=argparse.BooleanOptionalAction, default=False)
```
rknop: This looks like a reasonable way to do it.  I might add a help string saying what this does, and put in lots of warnings about how scary it is.  Or, maybe, just put in comments.

rknop (approved):
Looks fine.

Either fix the hardcoding of the `ltcv_provenance_key` here or make an issue that you need to fix that.  (See comment I made on the relevant line.)

Approving, under the assumption you will go the issue path.  (This will need to get fixed soon, i.e. before the end of next week.)

Maybe one or two other small comments to look at.

PR #210, opened 2025-11-24
Parallelizing campari.

ColeFMeldorf:
In this PR, we parallelize campari to improve speed. To do so, we parallelize in two places. Once when fetching the images and making cutouts, and again when building the models for each image. To do so, each of the code sections that were responsible for these were extracted into functions that could be called in a loop by `Pool` to run in parallel. These are `construct_one_image` and `build_model_for_one_image`. Tests have been added for both of these functions. I also added the nprocs argument.
Note that the regression values for the test lightcurves have slightly changed. This is because when called in parallel, the WCS objects are pickled, causing them to be slightly different. However, the difference in final flux is 1/1000 of a count, or less than 1/100 of a millimag, so I think it's safely ignorable.

### DIFF
--- a/changes/210.campari.rst
+++ b/changes/210.campari.rst
@@ -1,1 +1,1 @@
-Allowed for campari to run in parallel
+Allowed for campari to run in parallel using multiprocessing.


campari/campari_runner.py
```diff
@@ -272,7 +273,8 @@ def create_sim_param_grid(self):
         # self.param_grid = np.array(nd_grid, dtype=float).reshape(len(params), -1)
         # SNLogger.debug("Created a grid of simulation parameters with a total of"
         #                f" {self.param_grid.shape[1]} combinations.")
-        # self.diaobject_name = self.diaobject_name * self.param_grid.shape[1]  # Repeat the SNID for each combination of parameters
+        # self.diaobject_name = self.diaobject_name * self.param_grid.shape[1]
+        #  # Repeat the SNID for each combination of parameters
```
wmwv: What is this change?  Why is it part of this PR?

campari/data_construction.py
```diff
@@ -5,6 +5,7 @@
 from astropy.utils.exceptions import AstropyWarning
 from erfa import ErfaWarning
 import numpy as np
+from multiprocessing import Pool
```
wmwv: Alphabetize

campari/data_construction.py
```diff
@@ -24,7 +25,7 @@
 huge_value = 1e32
 
 
-def construct_images(image_list, diaobj, size, subtract_background=True):
+def construct_images(image_list, diaobj, size, subtract_background=True, nprocs=None):
```
wmwv: I suggest `nprocs=1` would be a reasonable default.

campari/run_one_object.py
```diff
@@ -150,101 +153,77 @@ def run_one_object(diaobj=None, object_type=None, image_list=None, size=None, ba
...
-
-            sn_matrix.append(psf_source_array)
-
-        elif sn_index >= 0 and prebuilt_sn_matrix is not None:
-            SNLogger.debug("Using prebuilt SN matrix for transient model")
+    model_results = []
+    if nprocs > 1:
+        with Pool(nprocs) as pool:
+            for i, image in enumerate(image_list):
+                kwarg_dict = {"image": image,
```
wmwv: this kwarg dict is mostly static.  Construct it outside the `if nprocs` and then just make the small changes necessary as you loop through.  This will make it much easier to keep things consistent between the with Pool and single-processer runs.

campari/RomanASP.py
```diff
@@ -252,6 +252,8 @@ def main():
     parser.add_argument("--add-truth-to-lc", action=argparse.BooleanOptionalAction, default=False, help="If True, "
                         "will add the truth fluxes from ou2024 to the lightcurve output. Default False.")
 
+    parser.add_argument("--nprocs", type=int, default=10, help="Number of processes to use. Default 10.")
```
wmwv: I would have gone with the singular for the variable name: `nproc`
Is there a SLURM or Python Pool convention for this being plural `nprocs`?

campari/tests/test_campari.py
```diff
@@ -813,3 +817,83 @@ def test_calculate_surface_brightness():
...
+            np.testing.assert_allclose(cutout.data, reg_cutout.data, atol=1e-7)
+            np.testing.assert_allclose(cutout.noise, reg_cutout.noise, atol=1e-7)
+            np.testing.assert_array_equal(cutout.flags, reg_cutout.flags)
+            np.testing.assert_array_equal(cutout.get_wcs()._wcs.to_header(), reg_cutout.get_wcs()._wcs.to_header())
+
+
+def test_build_model_one_image():
+
+    with open(pathlib.Path(__file__).parent / "testdata/reg_ra_grid.pkl", "rb") as f:
+        ra_grid = pickle.load(f)
```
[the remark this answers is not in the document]
ColeFMeldorf: Repackaged!

campari/tests/test_campari.py
```diff
@@ -813,3 +817,83 @@ def test_calculate_surface_brightness():
...
+    with open(pathlib.Path(__file__).parent / "testdata/reg_test_cutouts.pkl" , "rb") as f:
+        reg_cutout_list = pickle.load(f)
+
+    ra = 7.551093401915147
+    dec = -44.80718106491529
+    size = 19
+    truth = "simple_model"
+    subtract_background = True
+
+    for nprocs in [10, 1]:
```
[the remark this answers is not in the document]
ColeFMeldorf: That's super useful, thank you!

campari/run_one_object.py
ColeFMeldorf: I have no idea why I didn't just do that before.

campari/data_construction.py
ColeFMeldorf: I was going based off Pool "P" but I think you're right that's not the standard.

campari/model_building.py
```diff
@@ -555,3 +559,125 @@ def make_contour_grid(img_obj, numlevels=None, percentiles=[0, 90, 98, 100], sub
...
+
+
+#        sn_matrix.append(psf_source_array)
+
+    elif sn_index >= 0 and prebuilt_sn_matrix is not None:
+        SNLogger.debug("Using prebuilt SN matrix for transient model")
+        psf_source_array = None
+    else:
+        psf_source_array = None
+
```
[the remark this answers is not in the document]
ColeFMeldorf: It was just to throw the additional debug when preloading the model, but I've now clarified the logic.

campari/model_building.py
```diff
@@ -555,3 +559,125 @@ def make_contour_grid(img_obj, numlevels=None, percentiles=[0, 90, 98, 100], sub
     ra_grid, dec_grid = wcs.pixel_to_world(yy, xx)
 
     return ra_grid, dec_grid
+
+
+def build_model_for_one_image(image=None, ra=None, dec=None, use_real_images=None, grid_type=None, ra_grid=None,
```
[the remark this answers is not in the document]
rknop: We talked about this elsewhere.  I think the eventual goal (in a future PR) is to move all simulator stuff to snappl (where some of it is already), and take all of it out of campari.  At that point, campari would not know whether it's running on real images or not, it would just have images to run on.  It would mean all these `if real` blocks would go away.  This is a better test anyway, because it means exactly the same code paths are being run with simulated images and real images.

PR #220, opened 2025-12-08
Moved provenance creation earlier in the code.

ColeFMeldorf:


campari/data_construction.py
```diff
@@ -144,7 +144,7 @@ def construct_one_image(indx=None, image=None, ra=None, dec=None, size=None, tru
             bg = calculate_background_level(imagedata)
         else:
             # or we can read it from the image header if it's available.
-            bg = image_cutout.get_fits_header()["SKY_MEAN"]
+            bg = image_cutout.get_fits_header()["SKY_MEAN"] if "SKY_MEAN" in image_cutout.get_fits_header() else 0
```
rknop: This is scary....

You don't need to fix this in this PR, but put in an issue to do real background subtraction.  `SKY_MEAN` is not a standard header (plus, we're not always going to be able to use `get_fits_header`), so just assuming we know the background is 0 if that keyword doesn't happen to show up is not right.

rknop (approved):
Approved, but I suggest removing one of your command-line arguments (that you aren't using anyway), and also suggest creating an issue.  Check it out.

campari/data_construction.py
ColeFMeldorf: This issue already exists! This was also just a hack to get tests to pass for this one PR, this code looks different in another PR that I am about to merge, though it still relies on fits headers, hence the issue. See Issue 217

PR #221, opened 2025-12-11
Varying gauss tests

ColeFMeldorf:
In this PR, we introduce tests with a varying Gaussian PSF that I implemented in snappl. The goal is to recreate all of the results that we got with the static gaussian, which I was able to do. I am keeping track of the pull distributions and the bias and scatter levels for each test in a separate google sheet which I can provide on request. 
Also, I am currently storing all of the test data in photometry_test_data and not pushing it, because the total size is now pushing 7.1 GB. I can perhaps delete some of the images as they are not being used, and I could delete some of the static gaussian tests since they have been superseded by varying gaussians, but is there a better way to store this? 

I next want to add more realistic galaxies and move onto OU24 PSFs, I just wanted to get these changes merged before this gets too divergent.

Three tests currently do not pass:
FAILED test_gausspsfs.py::test_noiseless_shifted_22mag_host - AssertionError: 
FAILED test_gausspsfs.py::test_noiseless_shifted_22mag_host_varying - AssertionError: 
FAILED test_gausspsfs.py::test_both_shifted_nohost_varying - AssertionError: 
However I am going to leave them for now because they are all superseded by more difficult tests. I think the noiseless case fails because it is hard to understand exactly what the error distribution should look like when it's noiseless, and the last fails because PSF photometry is inherently a little biased when just fitting to a point source, but I do want to come back to this.

This is what a run on very faint sne under a 22 mag host looks like right now. I thought I could perhaps do better, but masao thinks there's no bias. Including this to see what people think.
image

image

### DIFF
--- a/campari/tests/test_gausspsfs.py
+++ b/campari/tests/test_gausspsfs.py
@@ -15,6 +15,10 @@ from snappl.logger import SNLogger
 
 SNLogger.set_level("INFO")
 
+# Note that to run these tests one needs to have an up to date checkout of
+# photometry_test_data
+# https://github.com/Roman-Supernova-PIT/photometry_test_data
+
 
 imsize = 19
 base_cmd = [


campari/tests/testdata/saved_lc_file.ecsv
```diff
@@ -43,8 +43,13 @@
 # - {dec: -41.0}
 # - {band: Y106}
 # - {diaobject_position_id: null}
+<<<<<<< HEAD
```
wmwv: Resolve the merge conflicts in this and other ecsv files.

campari/tests/test_gausspsfs.py
```diff
@@ -200,12 +211,19 @@ def perform_gaussianity_checks(residuals_sigma):
     mu, sig = norm.fit(residuals_sigma)
     # If I am not mistaken, this is equivalent to checking that the residuals are unbiased at 3 sigma confidence.
     mu_atol = 3 / np.sqrt(len(residuals_sigma))
-    SNLogger.debug("Fitted residuals mu: " + str(mu) + ", sig: " + str(sig) + "\n mu tolerance: " + str(mu_atol))
+    SNLogger.info(inspect.stack()[0][3])
```
wmwv: Are these two lines going to be interpretable in the output?

campari/tests/test_gausspsfs.py
```diff
@@ -4,11 +4,15 @@
 import numpy as np
 from matplotlib import pyplot as plt
 from scipy.stats import norm, skewtest
+import pytest
```
wmwv: Alphabetize imports

campari/tests/test_gausspsfs.py
```diff
@@ -917,10 +936,168 @@ def test_both_aligned_22mag_host():
...
+    np.testing.assert_allclose(lc["flux"], flux, atol=1, rtol=1e-3)
+
+    # The 4.14 comes from the fact that we set an error floor of 1 on each pixel. If we did not, this error
+    # would be zero for the noiseless case, but this causes issues with the inverse variance in other cases so
+    # we keep it.
+    # The tolerance has to be higher here because the rotation and changing in PSF sizes.
+    np.testing.assert_allclose(lc["flux_err"], 4.14, rtol=1.1e-2)
+
+
+def test_skynoise_shifted_22mag_host():
```
wmwv: If you expect this test to fail, then mark it as skip or expected fail.

wmwv (changes requested):
1. Sort out storing tests and test data with @rknop  I recall some discussion on Slack but not the resolution.
2. Mark tests that fail with expected fail and note why.
3. Clean up merge conflicts

PR #243, opened 2026-04-01
Self made tests

ColeFMeldorf:
In this PR we make it so that some tests can generate their own data so that others can run them.

To do so we add a python script that can run Rob's image simulator, which is then called by the tests if the data does not already exist.

To increase understandability we have also added a significant amount of documentation to the test files as well as packaged a lot of the accuracy checking of the tests into a single function.


campari/plotting.py
```diff
@@ -243,7 +244,12 @@ def generate_diagnostic_plots(fileroot, imsize, plotname, ap_sums=None, ap_err=N
     plt.close()
 
     plt.clf()
-    lc["flux_err"] = np.sqrt(lc["flux_err"] ** 2 + err_fudge**2)
+    if lc["flux_err"] is not None and err_fudge is not None:
+        lc["flux_err"] = np.sqrt(lc["flux_err"] ** 2 + err_fudge**2)
+    elif lc["flux_err"] is None:
+        # Noiseless tests will have no error.
+        err_fudge = err_fudge if err_fudge is not None else 0
+        lc["flux_err"] = np.full_like(lc["flux"], err_fudge)
```
wmwv: err_fudge defaults to 0, not None.  It feels like these lines could be simplified assuming that err_fudge is a float, and 0 if not being used.

campari/run_one_object.py
```diff
@@ -261,7 +261,10 @@ def run_one_object(diaobj=None, object_type=None, image_list=None, size=None, ba
                        f"r1norm: {r1norm}")
 
     flux = X[-num_detect_images:] if num_detect_images > 0 else None
-    inv_cov = psf_matrix.T @ np.diag(wgt_matrix**2) @ psf_matrix
+
+    # This multiplication is less memory intensive than the old way
```
wmwv: "the old way" is not defined.  Either remove "the old way" or define.

examples/perlmutter/campari_config_test.yaml
```diff
@@ -1,3 +1,8 @@
+
+
+# preloads:
+#   - /snpit_env/configs/rknop_dev_container_config.yaml
+
```
wmwv: Why does this add a commented-out configuration?

wmwv (approved):
Approve, but if you want to rework the function to build dicts instead of building a string, I encourage you to consider it.

campari/run_one_object.py
ColeFMeldorf: check

campari/plotting.py
ColeFMeldorf: Check!

ColeFMeldorf:
>Approve, but if you want to rework the function to build dicts instead of building a string, I encourage you to consider it.

In the interest of getting stuff merged, I am merging this, but raising an issue and opening a new PR to fix up the image simulator to make it not use the CLI.

ColeFMeldorf:
See Issue #244

PR #245, opened 2026-04-20
Imsim test

ColeFMeldorf:
In this PR, because I am having some issues with the image simulator in other branches, I add a sanity check test for campari to run the image simulator script here.

To clarify, I run the snappl image simulator from campari in order to generate images that I can then run campari on. I wanted to add a regression test that ensures that the image simulator is not changing as other things are modified. I put this test in campari and not snappl because I also want to test the script that campari has that wraps around the image simulator to call it.
The fits files are the images generated by the image simulator. Each is only about 30x30 pix so it's not a lot of memory.

One misc fix to get tests to pass: the partial_overlap.txt file keeps getting deleted somehow, so I made the test just write it before it begins. It's just a single pointing/sca combo.


wmwv:
I don't understand the goals of this PR.  Could you point to an Issue or re-articulate here what the goals of the PR are so that I can see if I agree the code is doing that?

wmwv:
> : the partial_overlap.txt file keeps getting deleted somehow,

What do you expect to happen?  Is this a file that is supposed to live in the test data permanently, or is it just generated for each test?

ColeFMeldorf:
@wmwv 

>What do you expect to happen? Is this a file that is supposed to live in the test data permanently, or is it just generated for each test?

I think before it got deleted when I made some mistake merging branches. Now I have it generate itself because it was literally a one line file.

>I don't understand the goals of this PR. Could you point to an Issue or re-articulate here what the goals of the PR are so that I can see if I agree the code is doing that?

I added what I sent to you in slack into the description of this PR!

campari/tests/test_campari.py
```diff
@@ -927,3 +932,88 @@ def test_build_model_one_image():
...
+    func_name = "image_simulator_script_test"
+    run_name = func_name + f"seed{simulation_number}"
+
+    # This is the file spat out by the image simulator sayingthe locations of all the files.
+    imagelist_filename = test_data_path / f"image_list_{run_name}.txt"
+    # Make sure we delete it beforehand
+    imagelist_filename.unlink(missing_ok=True)
+
+    run_sim(
+        seed=simulation_number,  # Set seed for reproducibility, this is the seed that Cole started with.
```
wmwv: Like, in life?  Or do you mean this was the seed the reference simulated data was simulated with?

campari/tests/test_campari.py
```diff
@@ -927,3 +932,88 @@ def test_build_model_one_image():
...
+        run_name_base=func_name,
+        bulge_R=2,
+        bulge_n=3,
+        disk_R=4,
+        disk_n=1,  # Simulated Galaxy Params
+        test_data_path=test_data_path,
+    )
+
+    regression_path = test_data_path / "image_simulator_script_regression/image_simulator_script_testseed0/"
+    regression_images = sorted(regression_path.glob("*image.fits"))
```
wmwv: Why sorted?

campari/tests/test_campari.py
```diff
@@ -927,3 +932,88 @@ def test_build_model_one_image():
...
+
+    # Zip will stop as soon as the shorter of the two lists runs out,
+    # so we check that they are the same length first to avoid missing any mismatches.
+    assert len(regression_images) == len(generated_images), (
+        f"Expected {len(regression_images)} generated images to match regression data, "
+        f"but found {len(generated_images)}."
+    )
+    assert [p.name for p in regression_images] == [p.name for p in generated_images], (
+        "Generated image filenames do not match regression image filenames."
+    )
```
wmwv: I would suggest you rethink this to take the list of generated images here and then find the matching reference one for each generated images.  The `basename` is the same, and you know the reference directory, so you can just construct the filepaths and make sure they exist.  I think that would simplify a bunch of these lines here and above.

campari/tests/test_campari.py
```diff
@@ -927,3 +932,88 @@ def test_build_model_one_image():
...
+                plt.imshow(np.log10(np.abs(gen_data - reg_data)), origin="lower")
+                plt.colorbar(label="log10( |generated - regression| )")
+                debug_image_path = test_data_path / f"debug_{gen_img.stem}.pdf"
+                SNLogger.debug(f"Saving debug image to {debug_image_path}")
+                plt.savefig(debug_image_path)
+                raise AssertionError(e)
+
+    # To update regression, do:
+    # mv /scratch/campari/campari/tests/testdata/image_simulator_script_test
+    #  /scratch/campari/campari/tests/testdata/image_simulator_script_regression
```
wmwv, line 1012: I don't think this is exactly right.  do you mean 
```
mv /scratch/campari/campari/tests/testdata/image_simulator_script_test/*
/scratch/campari/campari/tests/testdata/image_simulator_script_regression
```

campari/tests/test_campari.py
```diff
@@ -927,3 +932,88 @@ def test_build_model_one_image():
...
+        "Generated image filenames do not match regression image filenames."
+    )
+
+    for reg_img, gen_img in zip(regression_images, generated_images):
+        with fits.open(reg_img) as reg_hdul, fits.open(gen_img) as gen_hdul:
+            try:
+                reg_data = reg_hdul[0].data
+                gen_data = gen_hdul[0].data
+                assert gen_img != reg_img, ("Regression image and generated image paths should not be "
+                                            "the same, check the test setup.")
```
wmwv: I would put this test before the `try`.  It's a simple string comparison.  If it fails, you don't need any debug plots. to figure out the issue.

campari/tests/test_campari.py
```diff
@@ -927,3 +932,88 @@ def test_build_model_one_image():
...
+    assert [p.name for p in regression_images] == [p.name for p in generated_images], (
+        "Generated image filenames do not match regression image filenames."
+    )
+
+    for reg_img, gen_img in zip(regression_images, generated_images):
+        with fits.open(reg_img) as reg_hdul, fits.open(gen_img) as gen_hdul:
+            try:
+                reg_data = reg_hdul[0].data
+                gen_data = gen_hdul[0].data
+                assert gen_img != reg_img, ("Regression image and generated image paths should not be "
```
wmwv: Has this ever been triggered?  If it is, then that's a probably with the test, so this assert is testing the test, right?  It's not testing the main campari code.

Also, this isn't a guarantee that the files are different.  Different paths can point to the same place.

wmwv (changes requested):
This looks largely fine and reasonable.

Major request
1. I suggest rethinking the construction of the matching lists.  It's a little confusing and detailed.  I think you can make it much simpler by just assuming that a reference image exists for each generated test image.  And then if it doesn't just throw the error there.  But you don't have to do all of this sorting, and length checking, and then not match.
2. Think about when you're writing files that you can safely write to a central place and when you should make sure tests are isolated can can (at least in principle) be run in parallel.

campari/tests/test_campari.py
ColeFMeldorf: Clarified!

campari/tests/test_campari.py
ColeFMeldorf: It hadn't been triggered, but I was just feeling paranoid for some reason. I can remove it!

ColeFMeldorf: Actually since your other comment implies you don't think it should be removed entirely, I didn't, and instead replaced it with the more robust `assert not os.path.samefile` outside of the try/except as per your suggestion.

campari/tests/test_campari.py
ColeFMeldorf: Necessary when we were zipping over the two lists to compare, but now that I am no longer doing that I've removed it.

campari/tests/test_campari.py
ColeFMeldorf: Good catch!

campari/tests/test_campari.py
ColeFMeldorf: Done!

campari/tests/test_campari.py
ColeFMeldorf: Let me know if what I have changed it to is clearer!

ColeFMeldorf:
>Think about when you're writing files that you can safely write to a central place and when you should make sure tests are isolated can can (at least in principle) be run in parallel.

When you say "central" do you mean outside of the repo? Because my intent was that all of the test data for campari (other than the test data shared w/ phrosty) lives in the `testdata ` directory. I could move that elsewhere, however.

Also sorry, I don't follow. How does the fact that the FITS files are within the campari repo make it parallelizable? Or did you mean something else?

campari/tests/test_campari.py
wmwv, line 934: Is this (A) fixed input data or (B) output data written by the test?  If (B) making sure it's unique per test and so doesn't collide.

wmwv:
Imagine running all of the difference tests in `campari` at the same time.  Would they step on each other?

wmwv (changes requested):
Sorry, here was dangling comment that didn't get included in my previous review.  This is the part where I was asking about where you are writing files in the tests.

campari/tests/test_campari.py
ColeFMeldorf: It's the directory from which test data is read in, but also when you generate your own data on the fly inside campari but using snappl's image simulator it is also the directory your results are written to. However, since each test has a unique run name, I believe there should not be a conflict?

campari/tests/test_campari.py
```diff
@@ -773,6 +774,10 @@ def test_handle_partial_overlap():
     #  we know we're really running this test!
     assert not curfile.exists()
 
+    # make a textfile
+    with open(pathlib.Path(__file__).parent / "testdata/partial_overlap.txt", "w") as f:
```
wmwv: I suggest considering making this an isolated temporary file with TempFile or something like that.  Otherwise tests can't be run in parallel safely.

ColeFMeldorf: check!

wmwv (approved):
LGTM

PR #250, opened 2026-04-22
Save blank lc files

ColeFMeldorf:
MWV pointed out that it would make more sense if campari saved a blank LC file rather than nothing if a lc has no flux.
That being said, the way snappl is set up, I can't save a completely blank lc. Thus, I've made it one row full of nans.
I've also added a test that runs this new functionality.


wmwv:
Is the issue that `snappl.lightcurve.Lightcurve` won't save an empty lightcurve?  If that's the case, then don't worry about trying to work around that in `campari`.

The goal of this Issue was to minimize user surprise.  Having no lightcurve file is surprising, but having a lightcurve file with a row of Nan (and a -1) would also be surprising and hard to diagnose.

Issue #217, opened 2025-12-02
Campari BG subtraction should not use fits header.


ColeFMeldorf:
Another issue to make (don't change this for this PR): if you are going to support background subtraction, you should do it better than just subtracting an assumed thing out of a FITS header.   (Indeed, you should avoid using `get_fits_header` wherever possible, because that will almost certainly not be supported for ASDF files!)

You can use things like photutils' background subtraction routines.

_Originally posted by @rknop in https://github.com/Roman-Supernova-PIT/campari/pull/193#discussion_r2581939859_

PR #253, opened 2026-05-15
Smdc updates

ColeFMeldorf:
In this PR, we make miscellaneous changes to get all of the tests running on SMDC! This involved:

- [x] git adding some image list files that I didn't realize were untracked
SMDC does not have all the truth data that NERSC does, which necessitated two changes:
- [x]  Tests whjch rely entirely on truth data being available are now marked with requires_truth, allowing them to be skipped on SMDC.
- [x] Tests which only partially rely on truth data (i.e. the regression tests) have been modified to make the checking of truth data optional.
- [x] The photometry_test_data is on different location on SMDC. However, as far as I know, there is no way to make the filepaths in text files take in the location of the test data from the config. So, I wrote a new function, ` redirect_photometry_test_data` which parses incoming filepaths when reading the image list. If the photometry_test_data is in that path, it modifies the path to the correct location using the config value.
- [x] Created '/examples/SMDC' and relevant files.


campari/tests/test_campari.py
```diff
@@ -72,7 +73,7 @@ def campari_test_data(cfg):
     return cfg.value("photometry.campari_io.test_data")
 
 
-def compare_lightcurves(lc1_path, lc2_path, overwrite_meta=False):
+def compare_lightcurves(lc1_path, lc2_path, overwrite_meta=False, check_truth = False):
```
wmwv: No space around "="

campari/tests/testdata/test_imagelists/test_gaussims_nohost_poissonnoiseonlyseed51.txt
```diff
@@ -0,0 +1,150 @@
+/scratch/photometry_test_data/simple_gaussian_test/OU24_psf_tests/nohost_poissonnoiseonlyseed51/nohost_poissonnoiseonlyseed51_60070.0
```
wmwv: Why are these `/scratch/photometry_test_data` while all the other ones are `/photometry_test_data`?

[edited]

wmwv (changes requested):
1. You can inspect the `photometry_test_data` package to learn where the test data.  I encourage you to use this instead of a config file setting.

```
>>> import photometry_test_data
>>> dir(photometry_test_data)
['__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__']
>>> photometry_test_data.__path__
_NamespacePath(['/photometry_test_data'])
```

I think I would then just check for cases where it's not `/photometry_test_data` and then replace in that case.

2. What do we really need from the `tds.yaml` file?  We should be planning to move away from this.  What's the minimum information needed in this YAML file?

campari/tests/testdata/test_imagelists/test_gaussims_nohost_poissonnoiseonlyseed51.txt
ColeFMeldorf: Sorry I don't see the difference in the two strings you included?

wmwv: Sorry, `/scratch/photometry_test_data` vs `/photometry_test_data`

ColeFMeldorf: The location of certain files changed overtime in the podman environment that Rob made. Some of these files were constructed before this change, and some after. I've been doing some very annoying symbolic linking to get around this in the meantime, but this change would render this unnecessary on top of making it work on SMDC.

wmwv: Can you make them all start with "/photometry_test_data"?

ColeFMeldorf: Sure, there will just be a ton of files added to this PR, fair warning!

wmwv:
Thanks for the Slack discussion.  Given that, I think it's fine for tests to require a more special environment or additional assumptions.  So I'm find if you just make the tests all say "/photometry_test_data" and just require that to exist (and thus you're running in a container that has mounted that).

For regular operation, photometry_test_data isn't needed, so we don't need to do this in as general of a way.

campari/tests/testdata/test_imagelists/test_gaussims_nohost_poissonnoiseonlyseed51.txt
wmwv: Wait, why?  I think I'm missing an understanding of all of the testing.  Why does this requested change increase the number of files added to this PR?

wmwv: I thought the point was that they all live in `/photometry_test_data`?

campari/tests/testdata/newly_built_lc.ecsv
```diff
@@ -17,34 +17,35 @@
 # - {name: y_cutout, datatype: float64}
 # - {name: sky_background, datatype: float64}
 # meta: !!omap
-# - {cutout_size: 19}
+# - {cutout_size: 11}
```
wmwv, line 20: Why is this a change?

campari/tests/testdata/newly_built_lc.ecsv
```diff
@@ -45,8 +45,8 @@
 # - {dec: -41.0}
 # - {band: Y106}
 # - {diaobject_position_id: null}
-# - {provenance_id: ba6125c2-5d21-d282-90c9-15284e423e15}
-# - {diaobject_id: b52ae388-36aa-4be7-aa17-b2a61780d750}
+# - {provenance_id: ae2c0c7a-2491-c54e-79a3-fe6d71f2c973}
```
rknop, line 48: Just a note -- we're going to have to think about all of this in the future.  Right now, you've got tests that use things that are in external databases, but we're going to want to move to having the tests use entirely a self-contained environment.

It *might* be worth keeping some of the tests that require external databases around.  At the very least, you'll want to do the `@pytest.skipif` things I mentioned on Slack.  But, also, it might be worth putting those in a directory other than `tests`, to indicate that they're useful things we want to test, but that aren't part of the standard automatically-run test suite.  (Reason: we're going to want to make the standard automatically-run test suite only work in the self-contained environment, because it's going to be changing the database, and we don't want automatically-run tests changing an external database!)

rknop: Nothing to do here for this PR, just a warning that there's some pretty big test refactoring coming in the future.

campari/image_simulator_run.py
```diff
@@ -60,13 +60,13 @@ def run_sim(
 
     if output_path is None:
         # come up with a better default path for this! This is just a temporary placeholder.
-        output_path = "/scratch/photometry_test_data/simple_gaussian_test"
+        output_path = "/home/photometry_test_data/simple_gaussian_test"
         SNLogger.debug(f"No output_path provided, using default {output_path}")
 
     if im_sim_path is None:
-        im_sim_path = "/scratch/snappl/snappl/image_simulator.py"
+        im_sim_path = "/home/snappl/snappl/image_simulator.py"
```
rknop, line 67: ...this is another thing that's going to change at some point because of our `packages` subdir from photphest.  Don't worry about it right now.

HOWEVER, we also want to rethink this.  I don't think we want to be running `python image_simulator.py` from a shell (using `subprocess.run` or whatever), but we want to `improt snappl.image_simulator` and call a function.  I'll have to look at `snappl.imagesimualtor` to see if it's like this already, but ideally things like this have a `main()` that doesn't do anything other than parse command-line arguments to build the function call.  Then, to use from python, you just do the function call.

Again, not for this PR, but maybe create yourself an issue that this is something we want to think about.

examples/perlmutter/campari_config_outside_podman.yaml
```diff
@@ -18,14 +18,25 @@ photometry:
     use_real_images: true
     save_model: true
     save_debug: true
+<<<<<<< HEAD
```
rknop: Gotta fix this -- this is the stuff that shows up when you have a merge conflict.

examples/perlmutter/campari_config_outside_podman.yaml
```diff
@@ -34,33 +45,61 @@ photometry:
...
+    noise: 50
+    single_grid_point: false
+    sim_galaxy_scale:
+    - 1.0
+    sim_galaxy_offset:
+    - .00001414 # sqrt(2) * 1e-5
+    base_pointing: 276
+    base_sca: 10
+    run_name: whatever
+<<<<<<< HEAD
```
rknop: Another merge conflict that didn't get resovled before the file was committed that you need to fix.

examples/SMDC/campari_config_test.yaml
```diff
@@ -0,0 +1,84 @@
+# # This is used for the Nov 2025 podman container:
+# preloads:
+#   - /snpit_env/configs/nov2025_container_config.yaml
+
+# # This is used for the ASDF podman container:
+# preloads:
+#   - ${SNPIT_DEFAULT_CONFIG}
+
+preloads:
```
rknop, line 9: Need to keep this as is for now, but once you go to later snappl, you can go back to using `${SNPIT_DEFAULT_COFNIG}`

(That is fixed in snappl PR #217.)

examples/SMDC/campari_config_test.yaml
```diff
@@ -0,0 +1,84 @@
...
+    sim_gal_ra_offset: 0.00001
+    single_grid_point: false
+    sim_galaxy_scale: 1.0
+    sim_galaxy_offset: .00001414 # sqrt(2) * 1e-5
+    base_pointing: 276
+    base_sca: 10
+    run_name: whatever
+
+  snappl:
+    simdex_server: https://roman-desc-simdex.rknop.net
```
rknop: Point back to `https://roman-desc-simdex.lbl.gov`

The rknop.net one will go offline "soon".

robs_bash_script.sh
```diff
@@ -0,0 +1,64 @@
...
+echo $n_jobs
+#file_len=$(wc -l < /rick/romanisim/STARS_GAIA.csv)
+file_len=5
+echo $file_len
+stars_per_job=$((file_len / n_jobs ))
+echo $stars_per_job
+
+
+
+singularity run --bind /home/rkessler/:/rick:ro \
```
rknop: See if you can refactor this to not have an explicit `singularity run`, but rather run `/data/snpit/env/singrun_smdc_ricksim.sh`.

May require some other fixing because, for instance, that env wont' have a `/rick` directory.

rknop (changes requested):

campari/tests/testdata/newly_built_lc.ecsv
ColeFMeldorf: Got it! I will live in fear of it.

campari/image_simulator_run.py
ColeFMeldorf: See issue #268 !

examples/SMDC/campari_config_test.yaml
ColeFMeldorf: Issue #269 !

examples/SMDC/campari_config_test.yaml
ColeFMeldorf: got it!

robs_bash_script.sh
ColeFMeldorf: I don't even think this is functional right now, but I'll make that switch so I remember to do it when I come back to this.

examples/perlmutter/campari_config_outside_podman.yaml
ColeFMeldorf: No Idea how I missed this, sorry!

ColeFMeldorf:
Ok Rob comments addressed!

wmwv (approved):
LGTM

rknop (approved):

PR #254, opened 2026-05-27
Added the ability to glob image files rather than pass an image list

ColeFMeldorf:
Added the ability to glob image files. From the help statement I wrote:

>An alternate argument to --img_list. If you are running on a single image, or all of
>your images are in the same directory, here you can just give a path. Campari will then glob
>for the images following the regular expression you provide and use all of those images. For
>example: /path/to/images/*.fits will run on all fits images in that directory.
>/path/to/my_cool_image.fits will run on just that one image. Note that img_path and img_list 
>are mutually exclusive, and if both are passed, an error will be raised.


campari/tests/test_runner.py
```diff
@@ -207,6 +209,35 @@ def test_get_exposures(cfg):
...
+    test_args.image_collection_basepath = "/photometry_test_data/ou2024/images/"
+    test_args.img_path = "/photometry_test_data/sample_asdf_data/*.asdf"
+    runner = campari_runner(**vars(test_args))
+    runner.get_exposures(diaobj=diaobj)
+    mjd_list = [im.mjd for im in runner.image_list]
+    compare_list = [60627.500299]
+    np.testing.assert_allclose(np.array(mjd_list), np.array(compare_list), rtol=1e-5, err_msg=
+    "The set of MJDs recovered from globbing does not match the expected set of MJDs.")
+
+    # Unfortunately, this doesn't work with FITS. This is because FITS files in snappl are
```
wmwv: What, specifically, is the "this" that doesn't work?

campari/campari_runner.py
```diff
@@ -449,10 +478,16 @@ def build_and_save_lightcurve(self, diaobj, lc_model):
...
         """Parse the image list file if provided."""
-        with open(self.img_list) as ifp:
-            img_list_lines = ifp.readlines()
-        img_list_lines = [line.strip() for line in img_list_lines if
-                          (len(line.strip()) > 0) and (line.strip()[0] != "#")]
+        if self.img_list is not None:
+            with open(self.img_list) as ifp:
+                img_list_lines = ifp.readlines()
+                img_list_lines = [line.strip() for line in img_list_lines if
+                            (len(line.strip()) > 0) and (line.strip()[0] != "#")]
```
wmwv, line 478: (I know these are old lines).  In the future, I would discourage you from writing your own file parsing loop.  Use numpy, or astropy or something to just read the simple file.   I know it seems overkill, but it's much easier to read and verify.

campari/campari_runner.py
```diff
@@ -252,7 +253,35 @@ def __call__(self):
 
     def get_exposures(self, diaobj):
         """Call the find_all_exposures function to get the exposures for the given RA, Dec, and time frame."""
-        if self.img_list is not None:
+
+        if self.img_list is not None and self.img_path is not None:
+            raise ValueError("Cannot provide both img_list and img_path. These are two different ways to provide a list"
+                             " of images to run on, and if both are provided, it is ambiguous which the user intends to"
+                             " use. Please choose one or the other.")
+
```
wmwv: Do this check at parsing time.

campari/campari_runner.py
```diff
@@ -252,7 +253,35 @@ def __call__(self):
...
+        if self.img_path is not None:
```
wmwv: Write two separate get_exposure, one for each case and dispatch based on what is defined.  This makes it easier to understand and test each function.

```
def get_exposures(self, diaobj):
    if self.img_list:
         return get_exposures_from_img_list(...)
    if self.img_path
         return g_exposures_from_img_path(...)

def get_exposures_from_img_list(...):
      ...

def get_exposures_from_img_path(...):
      ...
```

`get_exposures_from_img_list` and `get_exposures_from_img_path` can also now be class methods instead of object methods.

wmwv (changes requested):
1.  Improve the documentation and clarity about img_list vs img_path.  Consider updating names.  I might suggest 'img_glob'.
  a. I'm concerned this is going to be a little confusing about whether you are passing a string that will be globbed within the Python code vs a glob that is expanded on the command line.
  b. img_list implies it will be an argument that accepts a list, but if I read correctly it is actually expect a filename.
2. Write documentation of expected behavior for the following strings and check that it works."a/b"
"a/b/"
"/a/b/c/"
"/a/b/c/*.asdf"
"{A,B}*.{asdf,fits}"

campari/campari_runner.py
ColeFMeldorf: Sure! I'll raise a separate issue.

campari/campari_runner.py
ColeFMeldorf: I had a slightly different fix where this is now all moved inside parse_image_list, I think it is much more easy to read than what I had before, let me know what you think.

campari/campari_runner.py
```diff
@@ -1,5 +1,6 @@
 import glob
 import pathlib
+from pathlib import Path
```
wmwv: Pick one.

wmwv (approved):
Looks good.  Only nitpick is that you have both `import pathlib` and `from pathlib import Path` in one of the files now.  Pick one.

wmwv:
Why did the tracemalloc stuff get pushed to this branch?

PR #260, opened 2026-07-13
Campari Memory Management

ColeFMeldorf:
- Managed campari memory to sub ~1.2 GB per process
- Added memory reporting and saving
- Added a test that sanity checks speed and time performance on NERSC
- sca_x and sca_y are now manually passed around because we are no longer carrying around the whole huge roman images to save memory.


wmwv:
As a quick starting point, I agree with Copilot suggestions except for one:

In the np.testing.assert_allclose, Copilot is correct that the `msg` isn't done right, but incorrect in its suggested solution.  You wanted to have `err_msg=msg` as one of the argument to `np.testing.assert_allclose` .

campari/run_one_object.py
```diff
@@ -111,7 +114,7 @@ def run_one_object(diaobj=None, object_type=None, image_list=None, size=None, ba
                                                                      subtract_background_method=
                                                                      subtract_background_method,
                                                                      nprocs=nprocs)
-
+    # del image_list  # Save memory
```
wmwv, line 117: ?

campari/tests/test_performance.py
```diff
@@ -0,0 +1,72 @@
...
+        "--photometry-campari-psf-transient_class ou24PSF_slow "
+        "--save_model --image-collection ou2024 "
+        " --no-save-to-db"
+        " --diaobject-collection ou2024"
+        f" --nprocs {nprocs}"
+        " --photometry-campari-grid_options-gaussian_var 1000 "
+        "--photometry-campari-print_memory_usage "
+        "--photometry-campari-save_memory_file_name test_regression"
+    )
+    assert output == 0, "The test run on a SN failed. Check the logs"
```
wmwv, line 46: Why do you keep running tests out to the command line?  You're in Python, just run them in Python.

You only need a couple of tests to make sure you parse arguments correctly.

campari/utils.py
```diff
@@ -369,19 +376,60 @@ def calculate_local_surface_brightness(image_object_list, cutout_pix=2, pixel_sc
...
+                pass  # some other process already created the file + header
+
+            # Append this process's row. O_APPEND makes each write land at the
+            # end of the file atomically, so rows from different processes
+            # don't get scrambled together even if they write at the same time.
+            fd = os.open(csv_path, os.O_APPEND | os.O_WRONLY)
+            with os.fdopen(fd, "a", newline="") as f:
+                csv.writer(f).writerow(
+                    [f"{elapsed_s:.3f}", label, f"{mem_gb:.4f}", current_time, pid]
+                )
```
wmwv, line 435: What is this doing?  What are you trying to achieve?

wmwv (commented):
1. I don't understand the logging.  Figure out something different than manually inserting lines into a file.  STDOUT, STDERR, one file-per-process, something.  Why does "print memory" save to a file?
2. Write tests from Python in Python.  The very common structure of these tests of calling out to to the command line makes them harder to read, easier to mess up writing the test strings, requires you to run the tests from a certain directory, and adds overhead of starting up a new process.  Is there something that needs to be changed in the test-running setup to make it easier to just import the modules and run the code in Python.
3. Address Copilot's mostly good suggestions (see comment in Conversation).
4. Add documentation about what the print memory function does.

campari/tests/test_campari.py
```diff
@@ -100,8 +100,9 @@ def compare_lightcurves(lc1_path, lc2_path, overwrite_meta=False):
     SNLogger.debug("NEW LIGHTCURVE DATA:")
     SNLogger.debug(lc1)
 
+    failed_cols = []
```
wmwv, line 107: Document what you're about to do here in the tests.  I think it's something like "we're going to collect the list of columns that fail/are missing instead of just failing on the first one."  Write out whatever is the correct answer/description.

campari/campari_runner.py
```diff
@@ -188,6 +189,8 @@ def __call__(self):
         # ra=self.ra, dec=self.dec
         #    mjd_discovery_min=self.transient_start, mjd_discovery_max=self.transient_end
 
+        clear_memory_file()
```
wmwv: What does this do?  Does it make a file that exists empty, or does it delete a memory usage tracking file?

campari/io.py
```diff
@@ -84,14 +84,13 @@ def build_lightcurve(diaobj, lc_model, obj_pos_prov=None, dbclient=None, cam_pro
...
+    for i, img in enumerate(cutout_image_list):
         if img.mjd >= diaobj.mjd_start and img.mjd <= diaobj.mjd_end:
             data_dict["mjd"].append(img.mjd)
             data_dict["observation_id"].append(str(img.observation_id))
             data_dict["sca"].append(img.sca)
-            x, y = img.get_wcs().world_to_pixel(diaobj.ra, diaobj.dec)
-            data_dict["pix_x"].append(x)
-            data_dict["pix_y"].append(y)
+            data_dict["pix_x"].append(lc_model.sca_x_locations[i])
+            data_dict["pix_y"].append(lc_model.sca_y_locations[i])
```
wmwv, line 93: I kind of agree with Copilot here, but I'm also okay with this just crashing and failing if the coordinates don't exist for a given image (or the array is shorter than the length of the array).

campari/requirements.txt
```diff
@@ -8,4 +8,5 @@ git+https://github.com/matroxel/roman_imsim.git
 scipy
 scikit-learn
 requests
-h5py
\ No newline at end of file
+h5py
+psutils
```
wmwv: Lint.

campari/utils.py
```diff
@@ -54,6 +57,8 @@ def __init__(
         post_transient_images=None,
         sky_background=None,
         image_collection_prov=None,
+        sca_x_locations = None,
+        sca_y_locations = None
```
wmwv: No space around `=`

campari/utils.py
```diff
@@ -369,19 +376,60 @@ def calculate_local_surface_brightness(image_object_list, cutout_pix=2, pixel_sc
     return LSB
 
 
-def print_memory_usage_summary(flag):
+def clear_memory_file():
```
wmwv: This removes the file not clears it.  (The distinction is that the file doesn't exist after calling this function.  "Clear" would imply it still exists but is empty.)

campari/utils.py
```diff
@@ -369,19 +376,60 @@ def calculate_local_surface_brightness(image_object_list, cutout_pix=2, pixel_sc
...
+        print(f"[MEM] {label}: {mem_gb:.2f} GB")
+
+        debug_dir = cfg.value("photometry.campari_io.debug_dir")
+        if cfg.value("photometry.campari.save_memory_file_name") != "None":
+            csv_name = cfg.value("photometry.campari.save_memory_file_name")
+            csv_path = f"{debug_dir}/{csv_name}.csv"
+            global _printed_location
+            if not _printed_location:
+                SNLogger.debug(f"Saving memory usage to {csv_path}")
+                _printed_location = True
```
wmwv, line 411: What does this do?

campari/utils.py
```diff
@@ -369,19 +376,60 @@ def calculate_local_surface_brightness(image_object_list, cutout_pix=2, pixel_sc
...
+
+            # Atomically CREATE the file only if it doesn't exist. O_CREAT|O_EXCL
+            # is a single OS-level operation: the OS checks-and-creates as one
+            # uninterruptible step, so it's impossible for two processes to both
+            # "win" this race, no matter how their timing overlaps.
+            try:
+                fd = os.open(csv_path, os.O_CREAT | os.O_EXCL | os.O_WRONLY)
+                with os.fdopen(fd, "w", newline="") as f:
+                    csv.writer(f).writerow(
+                        ["elapsed_seconds", "label", "memory_gb", "timestamp", "pid"]
```
wmwv, line 423: This is just standard logging.  I would encourage you to use a logger instead of writing your own custom function to do this.

wmwv (changes requested):
1. Consider using SN_logger or some standard logging package to write to the memory tracking file.
2. Minor comments.

campari/campari_runner.py
ColeFMeldorf: Changed the name of this to delete_memory_file! It deletes a memory usage tracking file.

campari/utils.py
ColeFMeldorf: This prints the location of where memory is being saved the first time this code is run, but not every time.

campari/utils.py
ColeFMeldorf: Updated the name!

campari/utils.py
ColeFMeldorf: done!

campari/tests/test_campari.py
ColeFMeldorf: done!

campari/requirements.txt
ColeFMeldorf: check

campari/utils.py
rknop: This isn't standard logging -- he's writing a csv file for later machine processing.

wmwv (approved):
LGTM

PR #262, opened 2026-07-15
Decreasing cyclometric complexity

ColeFMeldorf:


wmwv:
What are the goals of this PR?  I understand broadly, but having a detailed problem/goal statement as either an Issue or in the PR submission would help me understand if this PR was achieving those goals.

ColeFMeldorf:
> What are the goals of this PR? I understand broadly, but having a detailed problem/goal statement as either an Issue or in the PR submission would help me understand if this PR was achieving those goals.

Ack yes, it's because this PR wasn't really done, I just needed to see if it passed the ruff checks so it wasn't a draft, and it now automatically requests your review I think. Sorry!

campari/campari_runner.py
```diff
@@ -276,6 +248,41 @@ def __call__(self):
...
+        """ Sanity check provided ra/dec against the DiaObject's ra/dec. If they are far apart, raise a warning. """
+
+        if self.ra is not None:
+            if np.fabs(self.ra - diaobj.ra) > 1. / 3600. / np.cos(diaobj.dec * np.pi / 180.):
+                SNLogger.warning(f"Given RA {self.ra} is far from DiaObject nominal RA {diaobj.ra}")
+            diaobj.ra = self.ra
+        if self.dec is not None:
+            if np.fabs(self.dec - diaobj.dec) > 1. / 3600.:
+                SNLogger.warning(f"Given Dec {self.dec} is far from DiaObject nominal Dec {diaobj.dec}")
+            diaobj.dec = self.dec
```
wmwv: This fails at the poles.

Use AstroPy SkyCoord or something to do the (ra, dec) difference distance calculation correctly.

campari/campari_runner.py
```diff
@@ -545,6 +516,44 @@ def parse_img_list(self):
...
+            SNLogger.info(f"Saving Image WCS headers to {debug_dir}")
+            if lc_model.cutout_image_list is not None:
+                if not isinstance(lc_model.cutout_image_list[0].get_wcs(), snappl.wcs.GWCS):
+                    for i, img in enumerate(lc_model.cutout_image_list):
+                        hdul.append(fits.ImageHDU(header=img.get_wcs().to_fits_header(), name="WCS" + str(i)))
+                    hdul = fits.HDUList(hdul)
+                    filepath = debug_dir / f"{fileroot}_wcs.fits"
+                    hdul.writeto(filepath, overwrite=True)
+                else:
+                    SNLogger.warning("WCS is an astropy GWCS, which cannot be saved to a fits header."
```
wmwv: FITS

campari/image_simulator_run.py
```diff
@@ -25,6 +25,57 @@ def write_image_list(output_path, run_dir, run_name, test_data_path):
...
 
+def _check_sim_setup(run_dir, output_path, im_sim_path, test_data_path,
+                     just_rotate, just_shift, images_aligned, run_name_base):
+    if run_dir is None:
+        run_dir = "OU24_psf_tests"
+        SNLogger.debug(f"No run_dir provided, using default {run_dir}")
+
+    if output_path is None:
+        # come up with a better default path for this! This is just a temporary placeholder.
+        output_path = "/scratch/photometry_test_data/simple_gaussian_test"
```
wmwv: Just pick something from the campari config temp directories.

campari/image_simulator_run.py
```diff
@@ -25,6 +25,57 @@ def write_image_list(output_path, run_dir, run_name, test_data_path):
...
+    test_data_path = pathlib.Path(test_data_path)
+    if just_rotate and not images_aligned:
+        raise ValueError("Cannot both just rotate and have images not aligned")
+    if just_shift and not images_aligned:
+        raise ValueError("Cannot both just shift and have images not aligned")
+    if just_rotate and just_shift:
+        raise ValueError("Cannot both just rotate and just shift")
+
+    if run_name_base is None:
+        raise ValueError("run_name_base must be provided")  # This is important to avoid accidentally overwriting data.
```
wmwv: Why.  What would happen if run_num_base name were not provided?

campari/image_simulator_run.py
```diff
@@ -54,29 +105,13 @@ def run_sim(
...
-        SNLogger.debug(f"No test_data_path provided, using default {test_data_path}")
-    if just_rotate:
-        assert images_aligned == True, "Cannot both just rotate and have images not aligned"
-    if just_shift:
-        assert images_aligned == True, "Cannot both just shift and have images not aligned"
-    assert not (just_rotate and just_shift), "Cannot both just rotate and just shift"
+    run_dir, output_path, im_sim_path, test_data_path = _check_sim_setup(
+        run_dir, output_path, im_sim_path, test_data_path,
+        just_rotate, just_shift, images_aligned, run_name_base
+    )
```
wmwv: I suggest using named kwargs once you get past a few arguments.  It's easy to get things out of order.

campari/model_building.py
```diff
@@ -668,3 +674,25 @@ def build_model_for_one_image(image=None, ra=None, dec=None, use_real_images=Non
 
     print_mem("Finished image model building")
     return background_model_array, psf_source_array
+
+
+def prep_initial_guess(make_initial_guess, num_nondetect_images, grid_type, cutout_image_list, ra_grid, dec_grid,
+              num_total_images, initial_flux_guess, psf_matrix, subtract_background_method):
+    if make_initial_guess and num_nondetect_images != 0 and grid_type != "none":
+        SNLogger.debug("Making initial guess for the model")
+        x0 = generate_guess(cutout_image_list[:num_nondetect_images],
```
wmwv, line 683: Assumes order of images and that templates are only before.

campari/run_one_object.py
```diff
@@ -199,43 +181,20 @@ def run_one_object(diaobj=None, object_type=None, image_list=None, size=None, ba
...
-            / f"psf_matrix_{galaxy_psfclass}_{diaobj.name}_{num_total_images}_images{psf_matrix.shape[1]}_points.npy"
-        np.save(psf_matrix_path, psf_matrix)
-
-        sn_matrix_path = pathlib.Path(Config.get().value("photometry.campari_io.debug_dir")) \
-            / f"sn_matrix_{sn_psfclass}_{diaobj.name}_{num_total_images}_images.npy"
-        np.save(sn_matrix_path, sn_matrix)
-
-        SNLogger.debug(f"Saved PSF matrix to {psf_matrix_path}")
-        SNLogger.debug(f"Saved SN matrix to {sn_matrix_path}")
+    save_model_if_requested(save_model, psf_matrix, sn_matrix, galaxy_psfclass, sn_psfclass, diaobj, num_total_images)
```
wmwv: kwargs

campari/run_one_object.py
```diff
@@ -258,12 +217,9 @@ def run_one_object(diaobj=None, object_type=None, image_list=None, size=None, ba
...
 
-    # These if statements can definitely be written more elegantly.
-    if not make_initial_guess:
-        x0test = np.zeros(psf_matrix.shape[1])
-
-    if subtract_background_method == "fit":
-        x0test = np.concatenate([x0test, np.zeros(num_total_images)], axis=0)
+    # Using the images, hazard an initial guess.
+    x0 = prep_initial_guess(make_initial_guess, num_nondetect_images, grid_type, cutout_image_list, ra_grid, dec_grid,
+                        num_total_images, initial_flux_guess, psf_matrix, subtract_background_method)
```
wmwv, line 209: kwargs

campari/run_one_object.py
```diff
@@ -312,7 +265,7 @@ def run_one_object(diaobj=None, object_type=None, image_list=None, size=None, ba
             galaxy_only_model_images=galaxy_only_model_images,
             LSB=LSB, best_fit_model_values=X, sca_x_locations = all_sca_xs,
             sca_y_locations = all_sca_ys,
-            cutout_image_list=cutout_image_list, galaxy_images=np.array(galaxy_images), noise_maps=np.array(noise_maps),
+            cutout_image_list=cutout_image_list, noise_maps=np.array(noise_maps),
```
wmwv: Consider `np.asarray` if you want a no-op if noise_maps is already an array.  If you want to explicitly copy then make this explicit.

campari/utils.py
```diff
@@ -166,7 +166,7 @@ def calculate_background_level(im):
     return mean
 
 
-def get_weights(images, ra, dec, gaussian_var=1000, cutoff=4, error_floor=1):
+def get_weights(images, ra, dec, weighting, gaussian_var=1000, cutoff=4, error_floor=1):
```
wmwv: I suggest consider something like `use_weight` to make it more clear that it's a boolean determining whether or not to weight rather than not an array of weights.

wmwv: An add to docstring.

campari/utils.py
```diff
@@ -310,12 +313,28 @@ def calc_mag_and_err(flux, sigma_flux, band, zp=None):
 
 
 def banner(text):
+    """ Print a banner.
```
wmwv: using SNLogger.

pyproject.toml
```diff
@@ -165,7 +168,9 @@ select = [ 'F', 'E101', 'E111', 'E112', 'E113', 'E115', 'E117',
            'PLE0100', 'PLE0101', 'PLE0116',
            'NPY',
            'RUF018',
-           'E301', 'E302', 'E306', 'W505', 'D200', 'D212', 'RUF021', 'Q000']
+           'E301', 'E302', 'E306', 'W505', 'D200', 'D212', 'RUF021', 'Q000', 'C901']
+
+#            'D103'
```
wmwv: Remove or explain why commented out but not removed.

campari/utils.py
wmwv, line 220: Where does `5` come from?

changes/262.campari.rst
```diff
@@ -0,0 +1 @@
+Breaking up function into smaller functions.
```
wmwv: "functions" plural, or are you trying to say that this PR is focused on a particular function?

campari/utils.py
```diff
@@ -310,12 +313,28 @@ def calc_mag_and_err(flux, sigma_flux, band, zp=None):
...
+
+    Parameters
+    ----------
+    text : str
+        The text to print in the banner.
+
```
wmwv: Why extra line?

campari/plotting.py
```diff
@@ -242,106 +213,127 @@ def generate_diagnostic_plots(fileroot, imsize, plotname, ap_sums=None, ap_err=N
...
+    )
+
+    residuals = lc["flux"] - trueflux
+    window_size = 3
+    if len(residuals) >= window_size:
+        rolling_avg = np.convolve(residuals, np.ones(window_size) / window_size, mode="valid")
+        plt.plot(lc["mjd"][window_size - 1 :], rolling_avg, label="Rolling Average", color="orange")
+
+    if ap_sums is not None and ap_err is not None:
+        SNLogger.debug(f"aperture phot std: {np.std(np.array(ap_sums) - trueflux)}")
```
wmwv: np.asarray?

campari/plotting.py
```diff
@@ -242,106 +213,127 @@ def generate_diagnostic_plots(fileroot, imsize, plotname, ap_sums=None, ap_err=N
...
         plt.errorbar(
             lc["mjd"],
-            lc["flux"] - trueflux,
-            yerr=lc["flux_err"],
+            np.array(ap_sums) - trueflux,
```
wmwv: np.asarray?

campari/plotting.py
```diff
@@ -242,106 +213,127 @@ def generate_diagnostic_plots(fileroot, imsize, plotname, ap_sums=None, ap_err=N
...
     SNLogger.debug("Now generating light curve diagnostics...")
     # Now plot a light curve
-    if trueflux is not None:
-        plt.subplot(2, 2, 1)
+    _plot_diagnostic_lc_with_truth_if_provided(lc, trueflux, err_fudge, plotname, ap_sums, ap_err, ims)
+
+    SNLogger.debug(f"Generated saved diagnostic plots to {debug_dir}/{plotname}.png")
+
+
+def _plot_diagnostic_lc_with_truth_if_provided(lc, trueflux, err_fudge, plotname, ap_sums=None, ap_err=None, ims=None):
```
wmwv: Add docstring.

campari/plotting.py
```diff
@@ -242,106 +213,127 @@ def generate_diagnostic_plots(fileroot, imsize, plotname, ap_sums=None, ap_err=N
...
-                yerr=np.sqrt(lc["flux_err"] ** 2 + np.array(ap_err) ** 2),
-                marker="o",
-                linestyle="None",
-                label="Campari - Aperture Phot",
-                color="green",
-            )
+        non_transient_images = lc.meta["post_transient_images"] + lc.meta["pre_transient_images"]
+        image_sums = [np.sum(ims[i + non_transient_images]) for i in range(ims.shape[0] - non_transient_images)]
+        plt.errorbar(
+            lc["mjd"], np.array(image_sums) - trueflux, yerr=0, marker="o", linestyle="None",
```
wmwv: Do you sometimes want image_sums as a list and sometimes as an array?  I suggest just creating it as an array in the first place.

campari/plotting.py
```diff
@@ -242,106 +213,127 @@ def generate_diagnostic_plots(fileroot, imsize, plotname, ap_sums=None, ap_err=N
...
+    plt.xlabel("Flux")
+    plt.ylabel("Pull")
+    bin_stds, _, _ = binned_statistic(lc["flux"], pull, statistic="std", bins=bins)
+    plt.errorbar(bin_centers, bin_means, yerr=bin_stds, fmt="o", color="red", label="Binned Mean Pull with Std Dev")
+    plt.legend()
 
-        mu, sig = norm.fit(pull)
-        plt.plot(x, norm.pdf(x, mu, sig), label=f"Fit: mu={mu:.2f}, sig={sig:.2f}", color="red")
-        plt.legend()
+    plt.savefig(f"/{debug_dir}/" + plotname + "_lc.png")
```
wmwv: Consider making the postfix a keyword argument.  I find this an easy way to switch between making PNG and PDF files.

campari/plotting.py
```diff
@@ -242,106 +213,127 @@ def generate_diagnostic_plots(fileroot, imsize, plotname, ap_sums=None, ap_err=N
...
-        plt.axhline(0, color="black", linestyle="--")
-        plt.xlabel("Flux")
-        plt.ylabel("Pull")
-        bin_stds, _, _ = binned_statistic(lc["flux"], pull, statistic="std", bins=bins)
-        plt.errorbar(bin_centers, bin_means, yerr=bin_stds, fmt="o", color="red", label="Binned Mean Pull with Std Dev")
-        plt.legend()
 
-        plt.savefig(f"/{debug_dir}/" + plotname + "_lc.png")
+def _fudge_errors(lc, err_fudge):
+    """ Add an additional error term in quadrature to the flux errors.
```
wmwv, line 342: Would 'add_error_floor' be an appropriate name for this function?

campari/plotting.py
```diff
@@ -242,106 +213,127 @@ def generate_diagnostic_plots(fileroot, imsize, plotname, ap_sums=None, ap_err=N
...
 
-    SNLogger.debug(f"Generated saved diagnostic plots to {debug_dir}/{plotname}.png")
+    Parameters
+    ----------
+    lc : astropy.table.Table
+        The light curve table containing the flux and flux_err columns.
+    err_fudge : float
```
wmwv: `err_floor` or `error_floor`?

campari/plotting.py
```diff
@@ -242,106 +213,127 @@ def generate_diagnostic_plots(fileroot, imsize, plotname, ap_sums=None, ap_err=N
...
+        The additional error term to add in quadrature to the flux errors.
+    """
+    if lc["flux_err"] is not None:
+        error = np.sqrt(lc["flux_err"] ** 2 + err_fudge**2)
+    else:
+        # Noiseless tests will have no error.
```
wmwv: will have no 'flux_err' column so we create a new error array directly.

campari/plotting.py
```diff
@@ -242,106 +213,127 @@ def generate_diagnostic_plots(fileroot, imsize, plotname, ap_sums=None, ap_err=N
...
+        error = np.full_like(lc["flux"], err_fudge)
+    return error
 
 
-def plot_cutouts(cutout_image_list, ra, dec, diaobj=None, ncols=5, output_path=None):
+def plot_cutouts_if_requested(cutout_image_list, ra, dec, diaobj=None, ncols=5, output_path=None):
```
wmwv: Why is "_if_requested" in this function name?  Does this function accept a boolean that plots or not?  That would be weird.

campari/plotting.py
```diff
@@ -451,3 +433,53 @@ def plot_cutouts(cutout_image_list, ra, dec, diaobj=None, ncols=5, output_path=N
...
+        Dec of the supernova in degrees.
+    i : int
+        Index of the cutout image (for logging purposes).
+    """
+    try:
+        wcs = image.get_wcs()
+        sn_x, sn_y = wcs.world_to_pixel(ra, dec)
+        ax.scatter(sn_x, sn_y, marker="+", color="red", s=100, linewidths=1.5,
+                    label="SN" if i == 0 else None)
+    except Exception:
```
wmwv: Be more limited in the Exceptions you're catching.  If it's specifically about the SN position projection, then catch just whatever gets raised for that.

campari/run_one_object.py
wmwv: Why is "_if_requested" in the function name?

campari/run_one_object.py
```diff
@@ -272,16 +228,13 @@ def run_one_object(diaobj=None, object_type=None, image_list=None, size=None, ba
...
-        lsqr = sp.linalg.lsqr(psf_matrix*wgt_matrix.reshape(-1, 1),
-                              images*wgt_matrix,  atol=1e-12,
-                              btol=1e-12, iter_lim=300000, conlim=1e10, x0=x0test)
-
-        X, istop, itn, r1norm = lsqr[:4]
-        SNLogger.debug(f"Stop Condition {istop}, iterations: {itn}," +
-                       f"r1norm: {r1norm}")
+    wgt_matrix = np.sqrt(wgt_matrix)
+    lsqr = sp.linalg.lsqr(psf_matrix*wgt_matrix.reshape(-1, 1),
+                            images*wgt_matrix,  atol=1e-12, x0=x0,
```
wmwv: Odd indentation here, extra space before atol.

wmwv:
> > What are the goals of this PR? I understand broadly, but having a detailed problem/goal statement as either an Issue or in the PR submission would help me understand if this PR was achieving those goals.
> 
> Ack yes, it's because this PR wasn't really done, I just needed to see if it passed the ruff checks so it wasn't a draft, and it now automatically requests your review I think. Sorry!

Do you know how to run `ruff` locally?

wmwv (changes requested):
Assorted comments.

1. What's the goal of the PR?
2. Make sure utility functions have complete docstrings
3. Use kwargs to clarify arguments once you have more than a few arguments.  Particularly for booleans if you're passing a bare `True` or `False`. 
4. Consider function names.  E.g., "_if_requested" is a weird thing to have in a function name.

ColeFMeldorf:
> > > What are the goals of this PR? I understand broadly, but having a detailed problem/goal statement as either an Issue or in the PR submission would help me understand if this PR was achieving those goals.
> > 
> > 
> > Ack yes, it's because this PR wasn't really done, I just needed to see if it passed the ruff checks so it wasn't a draft, and it now automatically requests your review I think. Sorry!
> 
> Do you know how to run `ruff` locally?

I do, but I was briefly having some discrepancies between local ruff and github actions ruff. I just hadn't committed a change ://.

wmwv:
With reference to the goals.  I think this PR is a good first step.  Smaller functions better, but it depends a little bit, of course.  I'm okay with a 2-page function if the flow makes sense. But, I'm particularly sensitive to 15-line for loops; I really prefer these to be functions so that they can be tested cleanly.

As you keep going toward the future, think if there's logic design refactoring that can be done?  There are some logic strings of how to combine multiple arguments and options that make sense in the context of the function but suggest that the design is overly complicated.

campari/campari_runner.py
ColeFMeldorf: check

campari/campari_runner.py
ColeFMeldorf: check

campari/image_simulator_run.py
ColeFMeldorf: done!

campari/image_simulator_run.py
ColeFMeldorf: Otherwise I wouldn't know what to filename to save the images with! I could make a default, but I was worried about things being overwritten.

campari/image_simulator_run.py
ColeFMeldorf: check

campari/model_building.py
ColeFMeldorf: Images are intentionally sorted that way before they get to this stage.  See lines 92 - 107 in run_one_object.py:
```
    num_total_images = len(image_list)
    transient_image_list = [a for a in image_list if a.mjd >= diaobj.mjd_start and a.mjd <= diaobj.mjd_end]
    num_detect_images = len(transient_image_list)

    no_transient_images = [a for a in image_list if a.mjd < diaobj.mjd_start or a.mjd > diaobj.mjd_end]

    transient_mjds = [a.mjd for a in transient_image_list]
    no_transient_mjds = [a.mjd for a in no_transient_images]
    transient_argsort = np.argsort(transient_mjds)
    no_transient_argsort = np.argsort(no_transient_mjds)

    transient_image_list = [transient_image_list[i] for i in transient_argsort]
    no_transient_images = [no_transient_images[i] for i in no_transient_argsort]

    image_list = no_transient_images + transient_image_list  # Non detection images first, then detection images,
    # but still sorted by MJD.
```

campari/run_one_object.py
ColeFMeldorf: done!

ColeFMeldorf: I've also raised issue #275 , which will prevent this in the future.

campari/run_one_object.py
ColeFMeldorf: done, I've also raised issue https://github.com/Roman-Supernova-PIT/campari/issues/275 , which will prevent this in the future.

campari/run_one_object.py
ColeFMeldorf: done!

campari/utils.py
ColeFMeldorf: Done, and changed in all the config files

campari/utils.py
ColeFMeldorf: check

pyproject.toml
ColeFMeldorf: Moved to issue #276.

campari/utils.py
ColeFMeldorf: See comment on line 222:
```
            # NOTE: This 5 is here because when I made this function I was
            # checking my work by plotting and the *5 made it easier to see. I
            # thought the overall normalization
            # of the weights did not matter. I was half right, they don't matter
            # for the flux but they do matter for the size of the errors. Therefore
            # there is some way that these weights are normalized, but I don't
            # know exactly how that should be yet. Online sources speaking about
            # weighted linear regression never seem to address normalization. TODO
```
This is a longstanding issue that I need to resolve. However, I never use the gaussian weighting really in tests or real analysis runs so this is never hit. I am debating removing it entirely.

changes/262.campari.rst
ColeFMeldorf: fixed

campari/utils.py
ColeFMeldorf: fixed

campari/plotting.py
ColeFMeldorf: check

campari/plotting.py
ColeFMeldorf: check

campari/plotting.py
ColeFMeldorf: check

campari/plotting.py
ColeFMeldorf: fixed!

campari/plotting.py
ColeFMeldorf: check

campari/plotting.py
ColeFMeldorf: Certainly!

campari/plotting.py
ColeFMeldorf: made into err_floor

campari/plotting.py
ColeFMeldorf: Added

campari/plotting.py
ColeFMeldorf: In the function we have 
```
    if not Config.get().value("photometry.campari.preplot_cutouts"):
        return
```
So yes, it doesn't plot if that option is turned off. This was my solution for not having 10,000 if statements in the parent function. I changed the name to make it clear this function wouldn't always plot. If you prefer, the if can be in the parent function, but then it will be more complex. I am aware this choice was a manifestation of Goodhart's law.

campari/plotting.py
ColeFMeldorf: fixed!

campari/run_one_object.py
ColeFMeldorf: Same reasoning as the previous function, again, I can move the if statement up a level if you prefer.

ColeFMeldorf:
All comments addressed @wmwv !

campari/plotting.py
```diff
@@ -289,7 +289,8 @@ def _plot_diagnostic_lc_with_truth_if_provided(lc, trueflux, err_floor, plotname
         )
 
         non_transient_images = lc.meta["post_transient_images"] + lc.meta["pre_transient_images"]
-        image_sums = np.array([np.sum(ims[i + non_transient_images]) for i in range(ims.shape[0] - non_transient_images)])
+        image_sums = np.array([np.sum(ims[i + non_transient_images]) for i
+            in range(ims.shape[0] - non_transient_images)])
```
wmwv: This seems a touch complicated

The following seems immediately better
```
image_sums = np.array([np.sum(ims[i]) for i in range(non_transient_images, ims.shape[0])])
```

but I think you can even rewrite this as the shorter, simpler, and easier to understand:
```
image_sums = np.sum(ims[non_transient_images:], axis=(1, 2))
```

campari/plotting.py
wmwv: Okay, that's very confusing, but I agree with your function name.

You can leave this for now, but I would definitely prefer an explicitly if when deciding to call the function.  Relying on the implicitly global config this way is confusing to trace through.  The fact that you otherwise have 10,000 if statements (I assume the really number is 20) is a statement about the design of the way plotting is twined into the code instead of plotting and visualization just being a separate thing.

In addition to make the flow easier to understand, there's a dependency aspect.  In principle `campari` should be able to run without `matplotlib`  Plotting is just for debugging or choice of output.  Perhaps we'll never actually do this, but we should.  One particularly example of why I'm sensitive to this for matplotlib, when you have a brand new matplotlib installation, you have to generate the font cache.  If we ran with a fresh image each time, then there will be a 10-30-second startup to just generate the font cache.  In practice we'll need to distribute images that already have the font cache generated, but it's an example of design for separation of calculation and visualization and part of what I was getting at in thinking about improving the design.

campari/image_simulator_run.py
wmwv: Okay, requiring 'run_name_base' is fine.

It's the explanatory comment that's confusing.  It implies that there would be a name collision; but the hypothetical of your default doesn't exist in the code so the comment is confusing because it doesn't refer to the code.  I think it would be clearer if you remove the comment.

campari/utils.py
wmwv: Put the 5 comment closer to the code.  (particularly relevant because when reviewing a PR, but default one just sees the changed lines and the lines immediately above and below, but not 8 lines away.

campari/run_one_object.py
wmwv: Yes, please.

campari/image_simulator_run.py
```diff
@@ -25,6 +26,58 @@ def write_image_list(output_path, run_dir, run_name, test_data_path):
...
+    if just_rotate and not images_aligned:
+        raise ValueError("Cannot both just rotate and have images not aligned")
+    if just_shift and not images_aligned:
+        raise ValueError("Cannot both just shift and have images not aligned")
+    if just_rotate and just_shift:
+        raise ValueError("Cannot both just rotate and just shift")
+
+    if run_name_base is None:
+        raise ValueError("run_name_base must be provided")  # This is important to avoid accidentally overwriting data.
+        # I want to make sure the user consciously chooses a name for the run.
```
wmwv, line 58: If `run_name_base` is required, then I suggest moving it before keyword arguments that are not required.

campari/image_simulator_run.py
```diff
@@ -25,6 +26,58 @@ def write_image_list(output_path, run_dir, run_name, test_data_path):
     SNLogger.debug(f"Finished writing image list to {filename}")
 
 
+def _check_sim_setup(run_dir=None, output_path=None, im_sim_path=None, test_data_path=None,
+                     just_rotate=None, just_shift=None, images_aligned=None, run_name_base=None):
```
wmwv: Are just_rotate, just_shift, and images_aligned booleans?  If so, their defaults should be False or True.  If you mean to distinguish between False, True, and non set (None), that would be fine, but the code doesn't actually do that.  I think it would be clearer if you indicate the defaults here.

campari/image_simulator_run.py
wmwv: So I think that means it needs to be the _first_ argument, and should have no default.

wmwv (changes requested):
Some more comments regarding required arguments not being the last kwarg with a default, some more advanced notes about how np.sum works, and an encouragement to yes, write out all of the if statements and when that's annoying that's a design hint.

campari/plotting.py
ColeFMeldorf: Yes that's much better.

campari/image_simulator_run.py
ColeFMeldorf: done!

campari/image_simulator_run.py
ColeFMeldorf: Sounds good, I have them default to False to give the more realistic behavior by default.

wmwv:
You don't have to play catchup.  Unless you need a function from a newer version, you can develop against a particular snappl version tag, put that in the requirements (pyproject.toml) and make your life simpler.

ColeFMeldorf:
Kinda forgot people could read my git commit messages...

ColeFMeldorf:
Ok all comments addressed and all tests (except 2, which I am going to do separately) are passing

PR #267, opened 2026-08-18
All commented out code removed!

### LINKED ISSUE #240: Add ERA001 to ruff linter and get rid of all commented out code
See title

ColeFMeldorf:
For clarity. Also, a bug I just faced would have been found quicker if I had realized I had code I needed commented out.

### DIFF
--- a/campari/campari_runner.py
+++ b/campari/campari_runner.py
@@ -183,11 +183,6 @@ class campari_runner:
         """Run the Campari pipeline."""
 
         banner(f"Running SN {self.diaobject_name}")
-
-        # These will need to be re included once Issue #93 is resolved.
-        # ra=self.ra, dec=self.dec
-        #    mjd_discovery_min=self.transient_start, mjd_discovery_max=self.transient_end
-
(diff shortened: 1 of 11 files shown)


campari/campari_runner.py
```diff
@@ -184,13 +184,6 @@ def __call__(self):
         """Run the Campari pipeline."""
 
         banner(f"Running SN {self.diaobject_name}")
-
-        # These will need to be re included once Issue #93 is resolved.
-        # ra=self.ra, dec=self.dec
-        #    mjd_discovery_min=self.transient_start, mjd_discovery_max=self.transient_end
-
-        delete_memory_file()
```
wmwv, line 192: Is this also intended to be in this PR?

wmwv (commented):
There's one change in here that is not described by the PR.

Do you mean to remove `delete_memory_file` from the code and imports?  Totally fine if that's the behavior you want, but I didn't expect to see it in this PR and so am checking.

campari/campari_runner.py
ColeFMeldorf: No actually I think this was a mistake, thank you for catching that!

wmwv (approved):
LGTM

